Issue #10, opened 2019-02-28
Non-hex numbers ending with `f` missing unit tag


jeff-hykin:
<img width="121" alt="screen shot 2019-02-28 at 6 50 13 am" src="https://user-images.githubusercontent.com/17692058/53567538-20b3e500-3b25-11e9-88da-94ef0e3f7bc1.png">

The `f` should be tagged the same as the `l`.

PR #13, opened 2019-03-08
fix numeric literal suffixes #10

matter123:


jeff-hykin:
screen shot 2019-03-08 at 8 50 05 am

I haven't got a chance to fully look at the changes, however the current one just messes up on `0.0` which is kind of an edge case.
From what I can tell these are good changes in general though!

matter123:
The PR is for correctly marking unit suffix's on floating point numbers, e.g. `1e10f` as mentioned in #10. Several other minor issues regarding the decimal were also fixed.

Fixing `.314e6f` and `0xaBcDp5l` require modifying `numeric_pattern` to allow an initial decimal.

 1.4.11 | With Pull Request
------|-------------------
screen shot 2019-03-08 at 10 27 21 am | screen shot 2019-03-08 at 10 26 38 am

jeff-hykin:
Excellent.
That also fixes a lot of the literal syntax:

 1.4.11 | With Pull Request
------|-------------------
Screen Shot 2019-03-08 at 1 54 23 PM| Screen Shot 2019-03-08 at 1 55 06 PM

However it does change the markup of the single quote separators and I think it leaves out the - and + for the exponent `10e-29` but the single quote separators need to be changed anyways though. So I'm going to go ahead and merge this and then push a fix for the single quotes and - +.

jeff-hykin:
Alright the changes are live in version 1.4.12

jeff-hykin:
I reduced the `numeric_pattern` down to just the following:
```
possible_type_endings = maybe(/L|l|UL|ul|u|U|F|f|ll|LL|ull|ULL/)
valid_numeric_start = lookBehindToAvoid(@standard_character).then(/[0-9\.]/)
valid_numeric_middle = /[0-9\.xa-fA-Fp\+\-\']/
valid_numeric_pre_ending = /[0-9a-fA-F\.]/
numeric_pattern = valid_numeric_start.then(
    zeroOrMoreOf(valid_numeric_middle)
).lookBehindFor(valid_numeric_pre_ending).then(
    possible_type_ending
s).zeroOrMoreOf(@standard_character)
```
Feel free to edit it as it should be much easier to deal with now.

One of the only cases that still screws up is the `0xAB.cdpfL`

matter123:
Why do you have `possible_type_endings`? Isn't that sufficiently handled `zeroOrMoreOf(@standard_character)`? If not you're missing some cases as `u` & `l` can be in either order and with either capitalization.

`0xAB.cdpfL` is not a valid floating point literal even on hex floats the exponent is decimal. If you have a decimal exponent it seems to mark fine.

jeff-hykin:
Ah you're right, I didn't realize the exponent was limited to decimal values. Great every case should work then.

I realized we forgot the -+ for the hex floating point literal so I added that just now.
`0x0.234985p+10L`
`0x139804.234985p-10L`

I was thinking of keeping the `possible_type_endings` to add more specificity (label those endings as storage types and the others as user-defined types). Thanks for pointing out the missing cases, the pattern was copied and pasted from the atom/language-c pattern so I never checked it.

jeff-hykin:
Hey @matter123, just wanted to let you know I've been working on a refactor so if you were thinking about doing a pull request, you might want to wait 1 or 2 days until I can get the refactor merged in.

PR #21, opened 2019-03-15
Minor fixes

matter123:
This pull request fixes 3 minor issues.

1. The grammar only allowed spaces between the access control specifier and the colon rather than whitespace.
2. The integer suffix regex did not catch all integer suffixes (namely `ul` and `lu`)
3. Access control specifiers had the TM scope of `storage.type.modifier.access.control.$1.cpp` due to a non-capturing group.

With regards to point 3, my solution does not feel very idiomatic, is there a better way of marking a group as capturing without tagging it again?


jeff-hykin:
The whitespace change and the integer suffix look good. I can't believe I messed up the integer suffix after looking at the C++ spec, thanks for catching that.

As for the capture group `$1` I actually created a solution for that today that I'm excited to share!

For a pattern:
```ruby
thing = newPattern(
    match: /pattern1/,
    tag_as: "thing.$1"
).then(
    match: /pattern2/,
    tag_as: "thing.$1"
)
```

If you want to use itself in the tag name, you can now use `$match`
```ruby
thing = newPattern(
    match: /pattern1/,
    tag_as: "thing.$match" # this will be replaced with the result of pattern1
).then(
    match: /pattern2/,
    tag_as: "thing.$1"
)
```

If you want to use the result of a different group (e.g. pattern1 inside pattern2's tag) then you can create a named reference to it
```ruby
thing = newPattern(
    reference: "patternOne",
    match: /pattern1/,
    tag_as: "thing.$match" # the $match will be replaced with 'pattern1'
).then(
    match: /pattern2/,
    tag_as: "thing.$reference(patternOne)"  # the $reference() will be replaced with 'pattern1'
)
```

It will automatically add a capture group and it will calculate which group number is needed for the $match and $reference()

PR #38, opened 2019-03-19
fix #37

### LINKED ISSUE #37: syntax highlighting breaks after [ ] [ ]
image
syntax highlighting breaks after [ ] [ ]
it works only in ( )
it works again after first {

matter123:
The lambada `start_pattern` allowed a lambda to directly follow a closing square bracket.
As far as I know `[][](){}` could not be a valid sequence, so this should not break any valid code.


matter123:
The most recent commit fixes a similar issue where `test()[] = 5;` would also disable syntax highlighting until `}`. The only valid tokens after a capture list are `<`, `(`, or `{`. This commit checks that a lambda capture list is followed by one of those tokens or the end of the line

jeff-hykin:
Thanks for fixing this so quick

PR #49, opened 2019-04-01
C++ Fix for #47

### LINKED ISSUE #47: enum types missing highlighting
From: #45
Effects: C, C++

In the declaration `enum bar b;`, `bar` has incorrect highlighting. `bar` has the scope of `source.{c,cpp}`, it should have the scope `entity.name.type.{c,cpp} source.{c,cpp}`

Image:
C++

Example:
```C++
enum bar b;
int foo() {
    enum bar b;
}
```

matter123:
This provides a C++ fix for #47.

Additionally, this improves highlighting of inherited base classes. Previously only 1 base class per access specifier was allowed. This PR extends that to any number of base classes.

1.6.6 | This PR
-----|--------
Screen Shot 2019-03-31 at 7 50 49 PM | Screen Shot 2019-03-31 at 7 52 21 PM

This does not fix the same issue in the C grammar.


jeff-hykin:
Yeah for your original one you're right it's not needed. I abstracted out the `inhertance_context` though so it could be used in two different places, and the second place is when the complex regex is needed.

Right now:

With the complicated regex:
Screen Shot 2019-04-01 at 2 02 45 PM

Without the complicated regex:
Screen Shot 2019-04-01 at 2 03 31 PM

jeff-hykin:
The reason I did the abstraction is because your pattern matching was better, and it could be used in 2nd place.

matter123:
Ah, okay. You might still want to fix the regex tagging the space before the inherited class, however.

Issue #55, opened 2019-04-06
enum bug: punctuation.inhertance.colon.cpp


peaceshi:
set:
![image](https://user-images.githubusercontent.com/10380828/55676395-42b03c80-5907-11e9-9364-4c2fcfbe8528.png)
then
![image](https://user-images.githubusercontent.com/10380828/55676399-59569380-5907-11e9-9072-1879c3c597f8.png)
![image](https://user-images.githubusercontent.com/10380828/55676401-64112880-5907-11e9-9212-fe3373d13049.png)

matter123:
Can you provide some example code?

peaceshi:
> Can you provide some example code?

https://github.com/lhmouse/asteria/blob/master/asteria/src/syntax/statement.hpp

matter123:
well `punctuation.inheritance.colon` is probably better as `punctuation.type-specifier.colon` (@jeff-hykin ?).

I am not sure that the colon selected in the hover is an issue with this grammar. enum names are unscoped [decl.enum]/1. vscode-cpptools is doing something odd.

peaceshi:
> well `punctuation.inheritance.colon` is probably better as `punctuation.type-specifier.colon` (@jeff-hykin ?).tokenColorCustomizations` to hide the issue, see #7 for more information.
> 
> I am not sure that the colon selected in the hover is an issue with this grammar. enum names are unscoped [decl.enum]/1. vscode-cpptools is doing something odd.
> 
> You can use the setting `editor.

yeah, just I use `tokenColorCustomizations`, then discover this intesting bug.

matter123:
the slightly wrong scope is because #49 combined `class`,`struct`,`enum` and `union` all under one pattern, but `enum` uses the colon slightly differently.

matter123:
The scope should be more appropriate once #57 is accepted.

jeff-hykin:
I think `punctuation.type-specifier.colon` is good name. I'll try to get that merged in by tonight.

jeff-hykin:
This should be fixed by the next version, which should be pushed tonight. Let me know if it it's not fixed.

PR #54, opened 2019-04-05
WIP Fix #50

matter123:
1.6.8 | #54 
-------|-----
Screenshot from 2019-04-07 19-37-46 | Screenshot from 2019-04-07 19-36-35

Issue #48, opened 2019-04-01
generate_c.rb does not generate grammar


matter123:
`generate_c.rb` is not able to generate `syntaxes/c.tmLanguage.json` as It appears to be a stub and `c_grammar` is not a `Grammar` object. Additionally `generate_c.rb` is missing several referenced variables:
  - `function_definition_pattern`
  - `probably_a_parameter_2_groups`
  - `operator_overload_4_groups`
  - `variable_name_without_bounds`
  - `non_primitive_types`
  - `primitive_types`

This issue precludes fixing issues in the C grammar.

jeff-hykin:
Yeah, so far all of the changes have essentially been for C++, that file was created for the future. This issue should stay open until there is a proper c-generator file.

PR #57, opened 2019-04-07
Fix colon in enums #55

matter123:
enums are sufficiently different to deserve their own pattern separate from the class/struct/union pattern.
This PR does so and provides better scopes.


jeff-hykin:
I manually merged this one since there with conflicts caused by renaming and moving all of the files for #48 

I put the changes on this branch if you can take a look at them. They're already on master, but I'll hold of on publishing the extension until you think the changes are good.

matter123:
regressions:
 - anonymous structs are no longer tagged as `storage.type.struct`
 -     #ifdef __cplusplus
       extern "C" {
       #endif
is marked `invalid.illegal.stray-endif.cpp`

Overall fairly nice however

matter123:
Once this is corrected Ill redo #54.

jeff-hykin:
I believe this also had the problem before this:
```
#ifdef __cplusplus
class thing {
#endif
```
I got rid of the `invalid.illegal.stray`, but it meant the #endif is untagged. So I made a hacky fix for that since its part of the bigger issue of C++ if-directives in general

PR #65, opened 2019-04-09
change scope of new/delete Fix #64

### LINKED ISSUE #64: `new` changed from  keyword.control.cpp to keyword.operator.memory.new.cpp
It looks like new used to be keyword.control.cpp. Now it is keyword.operator.memory.new.cpp. Is this intentional? If it's intentional, we can add the it into the themes in VS Code, but ideally there wouldn't be too many changes like this since it requires constant updating of the themes.

matter123:
This PR changes the scope of new/delete to `keyword.other.memory.new` and `keyword.other.memory.delete`. To resolve the issue originally described in #64 

- `keyword.operator.memory.new` is not correct because `new` in new expressions is not an operator, only in `operator new(size_t)` is `new` an operator. Additionally many themes do not color operators.
- `keyword.control` is incorrect as `new` does not participate in control flow.

This PR (with default dark +)
Screenshot from 2019-04-09 08-50-11


jeff-hykin:
Thanks for looking this up, I didn't realize that was the case with the C++ spec. I've posted talking to alexr00 from the VS Code team, as soon she gives the okay I'll merge this in.

matter123:
Per https://github.com/jeff-hykin/cpp-textmate-grammar/issues/64#issuecomment-481341043 the scopes were changed to `keyword.operator.new` and `keyword.operator.delete`

jeff-hykin:
I went ahead and manually merged this in.

I also added a `memory` tag to all of the memory types
and a  `keyword.operator.wordlike` to all of the wordlike operators

This should work the best since it maintains backwards compatibility while still making it easy for every possible combination. For example:

`memory keyword.operator` will tag all of the memory keywords
`keyword.operator.wordlike keyword.operator` will tag all of the wordlike operators
`keyword.operator.new` will take priority over both of those.
`memory keyword.operator.new` will take priority over all three of the above.

Issue #66, opened 2019-04-09
capture references are off by one


matter123:
The pattern
```ruby
newPattern(
    match: /A/,
    tag_as: "a",
    reference: "part_a"
).maybe(@spaces).then(
    newPattern(
        match: /B/,
        tag_as: "$reference(part_a).b",
        reference: "part_b"
    )
).maybe(@spaces).then(
    newPattern(
        match: /C/,
        tag_as: "$reference(part_b).c",
        reference: "part_c"
    )
).maybe(@spaces).then(
    newPattern(
        match: /D/,
        tag_as: "$reference(part_c).d",
        reference: "part_d"
    )
)
```
generates the resultant grammar fragment
```yaml
- :match: "(A)\\s*(B)\\s*(C)\\s*(D)"
  :captures:
    '1':
      :name: a.test
    '2':
      :name: "$0.b.test"
    '3':
      :name: "$1.c.test"
    '4':
      :name: "$2.d.test"
```

Each capture reference is 1 less than it should be.

PR #67, opened 2019-04-09
Fix incorrect capture groups when using references #66

matter123:
The index for a reference in a pattern does not account for the fact that $0 is the whole pattern.


jeff-hykin:
I want to test this out in multiple scenarios.

There's an optimizer that converts 
`/(pattern)/` where group1 = `$1` 
into 
/pattern/ where group1 = `$0`

But it only activates under certain conditions. You may have already taken this into account so your fix is probably fine, it just might take me a bit to verify.

jeff-hykin:
I didn't fully test it, but this merge won't break any existing patterns so I'll go ahead and add it. If we find another off by one bug later we can fix it then.

PR #74, opened 2019-04-10
Fix #72 regressions caused by changes to blocks in 1.7.0

### LINKED ISSUE #72: Classes with opening brace on separate line breaks highlighting for rest of file
**VS Code** 1.33.0 
**Better C++ Syntax** 1.7.1
No other extensions installed and default settings

**Code:**
```cpp
class A
{
    void abc();
};
```

**Highlight:**
image

Return type `void` of method `abc` is not highlighted. If I add field it's also just white.

Does this extension require some initial setup (like coloring rules) or should I get nice highlighting just after installing it?

matter123:
1.7.0 caused some regressions to blocks, see #72 for more information.

This PR fixes the regressions.


jeff-hykin:
Didn't have a chance to look at these, but I'm going to merge and publish.

PR #92, opened 2019-04-12
allow comments before pre-processor #27

### LINKED ISSUE #27: comments before macros
https://github.com/atom/language-c/issues/159
```
/**/ #define a cout << "this works :/\n";
```

matter123:
This PR resolves #27.

This PR changes that start anchor of every pre-processor pattern to be either start-of-line or a look behind for `*/`


jeff-hykin:
This is a lot of changes, even though its a very simple change. I'd like to get a test this a bit before publishing. I also think we should leave this until we modernize the preprocessor section so that we have a better idea of how things work encase something gets broken.

matter123:
Yeah, I agree it's a lot of changes, it's also low-priority. So holding off is probably a good idea.

jeff-hykin:
I'm going to close this for now. Once `std_space` is made lets revisit it.

PR #98, opened 2019-04-15
Change to $self

matter123:
It might be best if you revert https://github.com/jeff-hykin/cpp-textmate-grammar/commit/898352f4943ffebb3f47157d03df57a90a33e2e8 and change it yourself. The cleanest way I found to do it, however, was to cherry-pick the changes since then back in.

Edit: Also this should be marked in #64, as at least one point I thin some grammars used the C grammar to extend their own. Edit: see http://textmate.1073791.n5.nabble.com/C-based-grammars-broken-for-embedding-restricting-base-reference-td25165.html


matter123:
I'm closing this PR.
 * it's several commits behind
 * this PR rewrites history
 * this PR breaks objective C syntax highlighting

PR #109, opened 2019-04-18
Fix #108

### LINKED ISSUE #108: Ternary operator messes with variable member highlighting
From https://github.com/Microsoft/vscode-cpptools/issues/3006
```
cfg.t_low_fan ? (cfg.t_low_fan == 1 ? "low" : "high") : "off";
```

matter123:
The ternary range had a limited selection of what was allowed between `?` and `:`. This PR adjusts the includes for the range to be `#function-call-innards`.

Screenshot from 2019-04-17 23-08-02


matter123:
Also @jeff-hykin why is the first group removed optimization not working?

jeff-hykin:
I had to remove it on ranges for some reason, but I'm not sure why.

jeff-hykin:
I think the weekend I'm going to upgrade the grammar along with looking into the objective-c $base/$self stuff so I'll try to look into it then

Issue #107, opened 2019-04-18
Fix objective-C and objective-C++ to not inherit


jeff-hykin:
Other languages (Objective-C, Objective-C++) include the C and C++ syntax

## What is the issue?
These other languages assume the C and C++ syntax use this thing `$base` that says "re-include every single pattern, and (if inside another language) only re-include all of that other languages patterns". Objective-C++ makes the assumption that basically every C++ namespace/struct/function/class/if/else etc uses `$base` inside of it.
```
// Objective-C++ code
int main() {
   // (this is the inside of a C++ pattern)
   // all Objective C++ patterns are allowed in here
}
```

## Why is this a bad assumption?
1. C and C++ (or any other language that uses $base) cannot work inside of markdown. One example of this is https://github.com/atom/language-c/issues/146 Here's why:
````markdown
This is a markdown file 
Let me give an example of C++ code
```cpp
class A {
   // because of $base, all the markup patterns are allowed here
   // (and C++ patterns are not allowed)
**bold** // marked as bold
int a = 10; // not highlighted at all
}
```
````
This is unfixable because of the Objective-C++ and Objective-C dependency.

2. Importing C++ is a bad assumption for Objective-C++ in particular because there are C++ constructs that are not allowed in Objective-C++ (Objective-C++ is not a superset of C++). Including the entire C++ language as the highest-priority group is requiring that the C++ syntax be incomplete for the Objective-C++ syntax to be correct.

3. `$base` is harmful even within C++ 
```
int main() {
  // all C++ patterns are allowed here
  // (not all C++ patterns should be allowed here)
  int main() {} // not a valid C++ function definition, but it gets highlighted because of `$base`
}
```
Because of the Objective C++ dependency, we are knowingly having to include incorrect patterns within C++, which is stopping us from adding support for major features like highlighting of variables and types.

## What is the solution?
It is easy to make the Objective-C++ syntax not depend on C++, by importing the valid pieces of C++ syntax into Objective-C++. This would mean C++ improvements would never affect Objective-C++, and it means we can finally start using C++ inside of markup.

jeff-hykin:
Once this is done we can clean up/remove the `cpp_base` pattern.

matter123:
For reference:
* the current markdown grammar is: https://github.com/Microsoft/vscode-markdown-tm-grammar
* the current Objective-C grammar is https://github.com/atom/language-objective-c

matter123:
[...]

PR #112, opened 2019-04-20
scope switch statements #110

### LINKED ISSUE #110: the colon of default case of switch missing parse.
image
maybe `punctuation.separator.default.cpp` useful.

matter123:
This fixes #110. `case` and `default` have been constrained to only match inside a switch statement.


jeff-hykin:
I'll try to get both branched merged by tomorrow night. Sadly haven't had a chance to work on anything yet/

peaceshi:
image
https://github.com/SDL-mirror/SDL/blob/master/src/video/windows/SDL_windowsevents.c#L88

matter123:
@peaceshi That's with the C grammar, I'll create a separate issue.

jeff-hykin:
I went ahead an manually merged this. I had to change it a bit to get all cases to work, adding `variableBounds[]` around the `switch`. The `()`'s were missing highlighting, I made them into a Range function and then had to modify `blockFinderFor` so that they would be matched even with no spaces (that part took forever). I'll wait to publish until you look over it though, the changes are in master.

I modified the way the grammar object works, and I'm using it now to import the switch statement into the c syntax. We can temporarily back-import until C++ is stable.

matter123:
The changes to switch look good to me. The rest of the changes you made today look fine. Renaming some of the scopes should probably end up on the log due to #107, but I don't think Objective-C++ grabbed a specific c++ scope as Objective-C did for C.

tristan957:
@peaceshi what are you doing in your comment to see the textmate grammar stuff?

matter123:
@tristan957 Open the command Palette (ctrl-shift-P or cmd-shift-p) `Developer: Inspect TM Scopes`

tristan957:
Ah thanks

Issue #113, opened 2019-04-21
punctuation matches between ( and ) after operators and template calls


pasha-kuznetsov:
```json
        {
            "scope": [
                "punctuation"
            ],
            "settings": {
                "foreground": "#FF0000",
                "fontStyle": "bold"
            }
        }
```

<img width="708" alt="Screen Shot 2019-04-21 at 11 54 21 AM" src="https://user-images.githubusercontent.com/14935558/56474283-91dc8e00-642c-11e9-96f8-9bb4ff76c921.png">

pasha-kuznetsov:
It's `punctuation.section.parens-c` defined in [generate.rb:1954](https://github.com/jeff-hykin/cpp-textmate-grammar/blob/master/cpp/generate.rb#L1954):
```rb
    "parens-c" => {
        name: "punctuation.section.parens-c",
        begin: "\\(",
        beginCaptures: {
            "0" => {
                name: "punctuation.section.parens.begin.bracket.round"
            }
        },
        end: "\\)",
        endCaptures: {
            "0" => {
                name: "punctuation.section.parens.end.bracket.round"
            }
        },
        patterns: [
            {
                include: "$base"
            }
        ]
    }
```

matter123:
Thanks

matter123:
Per https://github.com/jeff-hykin/cpp-textmate-grammar/pull/112#issuecomment-485226021, this will probably get fixed tonight.

PR #114, opened 2019-04-21
change scope of parens group #113

matter123:
The scope of a group of parenthesis was changed from `punctuation` to `meta` in both the C and C++ grammars.


matter123:
Closed via e66a2ca

PR #122, opened 2019-04-24
Number improvements

matter123:
From example.cpp
```c++
// invalid literals
1Ex
0xe+foo
0x1Px
// odd valid literals
0x'12345
0'12345
0x1Ex
1Px
```

Screenshot from 2019-04-23 17-45-34

This PR adds support for marking invalid numbers as `invalid.illegal.numeric.constant`, and for some slightly odd corner cases that were missed.

This PR unifies the C++ and C number handling. 

Edit:
`0xe+foo` is in fact invalid https://godbolt.org/z/CFM5V_ https://godbolt.org/z/P-ssZ1
and [[lex.pptoken]/4](http://eel.is/c++draft/lex#pptoken-4) (non normative):
> [ Example: The program fragment 0xe+foo is parsed as a preprocessing number token (one that is not a valid floating or integer literal token), even though a parse as three preprocessing tokens 0xe, +, and foo might produce a valid expression ...

though gcc suggest that operator""+foo would make the expression valid but that is not a valid ud-suffix


matter123:
For reference with 1.8.7:
Screenshot from 2019-04-23 18-09-46

jeff-hykin:
I might have to wait till this weekend to go over this and the alignas PR.

I like the extraction, thats probably the right way to go.

matter123:
The number mentioned in https://github.com/jeff-hykin/cpp-textmate-grammar/pull/13#issuecomment-471094341 `0xAB.cdpfl` is not currently marked as invalid because `*_digits` are just greedy, not possessive. The regex engine backtracks and parses it as `0x`, `AB.c`, and a udl of `dpfL` but that is not [valid](https://godbolt.org/z/6sbdf5) the standard requires that hex_digits consume as many characters as possible before considering UDL's

jeff-hykin:
What does `allow_udl` mean?

matter123:
Probably could be a better name, but is a C++ number which have user-defined literals, or a C number which do not.

jeff-hykin:
Oh okay 👍

jeff-hykin:
This was a great change, I renamed some stuff and removed the need for the `grammar` object requirement. I'm a little bit concerned about mixing the C and C++ stuff, but I think its find for now.

matter123:
This particular sharing should be okay as both C and C++ have the same grammar for `pp-number`
(C++ [[lex.ppnumber](http://eel.is/c++draft/lex.ppnumber)) (C11 N1570 A.1.9)

jeff-hykin:
Screen Shot 2019-05-04 at 3 45 49 PM
Minor regression with this change

jeff-hykin:
Fixed it by changing the pattern to
`start_pattern = lookBehindToAvoid(/\w/).lookAheadFor(/\d|\.\d/)`

But I think there's another minor issue
Screen Shot 2019-05-04 at 3 55 24 PM

matter123:
No.
The first green is on an `integer_literal` where `f` is not an `integer_suffix`
The second green is on a `floating_literal` where `LL` is not a `floating_suffix`

jeff-hykin:
okay great 👍 we should be all good then

matter123:
So despite that a decimal point is needed for a number to be considered a floating constant [[lex.fcon]#fractional-constant](http://eel.is/c%2B%2Bdraft/lex.fcon#nt:fractional-constant). and `f` is not a integer-suffix [[lex.icon]#integer-suffix](http://eel.is/c%2B%2Bdraft/lex.icon#nt:integer-suffix). 
pretty much all compilers compile `auto a = 1e10f;` https://godbolt.org/z/X64wm4. Ill submit a PR fixing that.

Issue #121, opened 2019-04-23
class inheritance highlighter gets confused by macros and comments


zrax:
If a comment or macro exists anywhere after the `class` keyword and before the last parent class, the highlighting of parent classes and access specifiers is lost.

Examples:
![image](https://user-images.githubusercontent.com/714191/56622236-1ae9f580-65e4-11e9-94f1-7c0e12fce9b5.png)

matter123:
With the first pattern, the grammar has no way of knowing that `DLLEXPORT` is actually __declspec(dllexport). `class __declspec(dllexport) ChildClass` should be highlighted correctly, however.

MonaTheMonad:
The `alignas` specifier also seems to break highlighting in a similar manner.

matter123:
The comment is fixed in 1.8.9.

jeff-hykin:
<img width="779" alt="Screen Shot 2019-06-24 at 6 40 24 PM" src="https://user-images.githubusercontent.com/17692058/60058841-941ecb00-96af-11e9-95bd-aac99fd7f0a6.png">
Alignas is fixed. `DLLEXPORT` isn't fixed, but it doesn't cause a cascading break anymore. Since DLLEXPORT isn't a built in I don't plan to support it directly. OS/Library specific things (within textmate) would be a best fixed by a grammar injection extension so that people can toggle it on and off.

PR #126, opened 2019-04-24
add alignas as an attribute type

matter123:
This fixes the `alignas` issue reported at https://github.com/jeff-hykin/cpp-textmate-grammar/issues/121#issuecomment-486286704.

There should probably be a refactor of inline_attribute soonish to support, support attributes like `PACKED` or `DLLEXPORT`


jeff-hykin:
Are spaces allowed in `__declspec(`? like `__declspec    (`

matter123:
Looks like yes https://godbolt.org/z/NqM5by

jeff-hykin:
okay, then I think we should take it out of the representation, or we should make the representation a regex that matches it. I'll try to find the best way to incorporate that for the other similar things too.

I think it might be best to remove the `()`'s from the representation, and then just have the `function-like` adjective to indicate that it needs `()`'s in the pattern

matter123:
Ok, the concern is with __attribute__(( as it has double parenthesis on both sides. e.g. `__attribute__((noreturn))` or `__attribute((constructor(101)))`

jeff-hykin:
I went ahead an manually merged this in just now. I moved them out of tokens though so that we can change them to include spaces in the future.

Issue #134, opened 2019-04-27
function parameter highlighting improvements


matter123:
List of possible paramater syntaxes
```yaml
function_parameters:
    1: Type (*name)(:function_parameters) = Value
    2: Type (*)(:function_parameters) = Value
    3: Type name = Value
    4: Type = Value
    5: Type (*name)(:function_parameters)
    6: Type (*)(:function_parameters)
    7: Type name
    8: Type
    9: <empty>
```

currently, the support is rather poor.
the variable is correctly identified in cases 1,3,5, and 7.
in cases 1,2,5, and 6, the return type of the function pointers is tagged as a function.
in case 4 the type is marked as the variable.
no cases mark the default value.

#71 fixes case 3 and 7.

matter123:
The eventual PR I'll submit Saturday or Sunday should include: this, function pointers, and #71 applied to class/struct definitions as all three are interrelated.

matter123:
From the `Explore-#71` branch
![Screenshot from 2019-04-26 20-02-14](https://user-images.githubusercontent.com/714007/56844044-f9487280-685e-11e9-9611-e6c60d2ce917.png)

matter123:
Is there any issue with adding support for comments, and string/numeric/language constants to the list of parameters and using it for both function definitions and function calls? It could be a generator so that inside a function foo(a,b,c) highlights a, b, and c as a variable but outside a function as a type.

jeff-hykin:
I think next time I work on the parameters I'll make a pattern for defaulted parameters.
If its a PatternRange, with a start of `=` and an end of lookAhead for `,` or `)` I think it can include all of the expresson_context patterns in a really solid way.

I think the pattern could also be applied to the template evaluative context too (e.g. `template<class B = C>`)

matter123:
Closing as this is fixed in master.

![Screenshot from 2019-06-22 19-42-25](https://user-images.githubusercontent.com/714007/59970886-e16f3100-9525-11e9-8d6e-8b1435dd7f00.png)

jeff-hykin:
Alright all those cases and more (whenever there was an array) were fixed.

Issue #71, opened 2019-04-10
Improve tagging for variable declarations


matter123:
To combine #29 #59 and #61 this issue is to track improving the tagging for variable.

Specifically tagging the patterns `A B;` and `A B{C};` with `A` as a type and `B` as a variable. `A B(C)` is excluded as `B` is ambiguous and `A` is already tagged as a type.

matter123:
@jeff-hykin are you aware of a statement that makes `A B` or `A B{C}` ambiguous?

jeff-hykin:
Well there are always macros which ruin absolutely everything, but thats kind of its own issue. 

I think there needs to be a well defined context before matching `A B` and `A B{C}`. Right now much of the code is free-form. The bodies of functions are not matched with the function name, `if` statements, `switch` statements, etc the keyword and bodies are matched independently, so we don't know much about the document structure.

For example, before the change last night with the class/struct/enum blocks, the tail of the enums wasn't matched.
```
enum Thing {
   a, b,c
}
enum1 enum2;
```
Before that `enum1 enum2;` would have been contextless, and would be matched with the general `A B` pattern.

If we can find a specific context though, and/or start actually providing more context structure to the document then it will probably be less of a problem to match `A B`. 

The `blockFinderFor()` was created so that we could start adding that kind of structure for functions, if statements, and anything else that uses `{}`'s

jeff-hykin:
In terms of the C++ spec, and assuming the right context, and no macros are involved, I don't know of an ambiguous case.

There are lots of edge cases I think, qualifiers like `extern`, reference/deference, scope resolution, templates, etc, but I don't know of anything that would make it unmatchable.

matter123:
``` c++
enum Thing {
   a, b,c
}
enum1 enum2;
```
isn't actually valid as multiple variable declarations need to be separated by commas, but I see your general point.

matter123:
I created the branch [Explore-#71](https://github.com/jeff-hykin/cpp-textmate-grammar/tree/Explore-%2371) to see how implementing this would interact with the rest of the grammar. currently only the pattern `A B` is supported. `A B{C}` is not excluded from being matched but there is no special support for `C`.

Images:
<img width="645" alt="Screen Shot 2019-04-26 at 11 14 04 AM" src="https://user-images.githubusercontent.com/714007/56827877-d98c5c80-6814-11e9-9329-a8009cc25fd3.png">
[...]

Issue #26, opened 2019-03-16
constructor gets confused by function pointers


jeff-hykin:
https://github.com/atom/language-c/issues/129

```
struct foo {
    int (*bar)();
};

union baz {
    int (*quz)();
};
```

int (*bar) is tagged as a constructor

matter123:
The `Explore-#71` branch contains a fix for this.
![Screenshot from 2019-04-26 20-04-30](https://user-images.githubusercontent.com/714007/56844025-a7075180-685e-11e9-8f53-7f1b300321cd.png)

```ruby
#
# Function pointers
#
    cpp_grammar[:function_pointer] = Range.new(
        start_pattern: qualified_type.maybe(@spaces).then(/\(/).maybe(@spaces).then(
                match: /\*/,
                tag_as: "variable.other.pointer.function",
            ).maybe(@spaces).maybe(
                match: identifier,
                tag_as: "variable.other.pointer.function"
            ).maybe(@spaces).zeroOrMoreOf(array_brackets).then(/\)/).maybe(@spaces).then(/\(/),
        end_pattern: /\)/.then(after_declaration),
        includes: [
            :function_parameters,
        ]
    )
```

Issue #127, opened 2019-04-24
posessive / lazy modifiers


matter123:
Because of the maximal munch rule [lex.pptoken/3.3](http://eel.is/c%2B%2Bdraft/lex.pptoken#3.3) The grammar needs a way to mark quantifiers as possessive or lazy.

I had some thoughts for what the ai would look like

1.
```ruby
possessive(oneOrMoreOf(/foo/))
```
2.
```ruby
oneOrMoreOf(/foo/).posessive()
```
3.
```ruby
oneOrMoreOf(match: /foo/, :posessive)
```
4.
```ruby
oneOrMoreOfPosessive(/foo/)
```

Advantages:
1 & 2. minimizes code repetition
3. cleanest looking
4. matches the pattern established by lookarounds

Disadvantages:
1 & 2. requires unwrapping the non-capturing group
2. confusing that `.posessive()` applies to the preceding element
3. requires specifying `match:`
4. create a total of 9 functions for qualifiers, or 12 if atomics are supported

@jeff-hykin thoughts?

matter123:
For reference number 2 is
```ruby
    def posessive()
        self_as_string = self.without_default_mode_modifiers
        index = self_as_string[-1] == ')' ? -2 : -1
        if not /[+*?}]/ =~ self_as_string[index]
            raise "\n\n attempted to make a non quantifier pattern posessive\npattern is #{self_as_string}"
        end
        new_regex = /#{self_as_string.insert(index, '+')}/
        
        new_regex.group_attributes = self.group_attributes
        return new_regex
    end
```
number 3 is
```ruby
modifier = ""
if attributes[:type] == nil or attributes[:type] == :greedy
    # pass
elsif attributes[:type] == :posessive
    modifier = "+"
elsif attributes[:type] == :lazy
    modifier = "?"
else
    raise "\n\n argument :type is not :posessive, :greedy :lazy\nregex: #{self.without_default_mode_modifiers}\arguments: #{attributes}"
end
attributes.delete(:type)
```
then interpolating modifer into maybe, oneOrMoreOf, and zeroOrMoreOf's constructed regexps

1 &4 should just be modifications upon those

matter123:
This is blocking some aspects of #135 as `:scope_resolution` can backtrack and cause the lookBehindToAvoids to succeed when it shouldn't.

jeff-hykin:
[...]

PR #135, opened 2019-04-27
Explore variable declares, func pointers, parameter improvements, and more

matter123:
This PR implements a fix for #26 and #134, as well as a partial fix for #71.

Patterns added:
 - `:qualified_type` a single type with an optional template call and preceding scope, avoids non-type tokens
- `:declaration` a qualified type followed by a variable name and zero or more square brackets, avoids preceding `(`
- `:declarations` same as declaration but allows multiple variables;
- `:type_alias` a `using` type alias (e.g. `using result_type = size_t;`)
- `:function_pointer` a function pointer declaration, optional name, any number of parameters.
- `:function_parameters` a single function parameter


matter123:
Screenshot from 2019-04-26 20-46-23

matter123:
this does not support yet multiple declarations where one of the variables is immediately succeeded by `{`.
e.g.
```c++
int a, b{5}`
```
Screenshot from 2019-04-26 20-55-37

note how 5 has no syntax highlighting.

matter123:
Screenshot from 2019-04-27 09-52-59

Fixed the issue with no comment highlighting and the lack of univeral initalization highlighting.

matter123:
currently, the pattern causes the union scope not to end.

matter123:
This is currently unsuitable for merging.

Regression:
-  functions with multiple parameters
- asm blocks
 - TODO

matter123:
function definitions are particularly troublesome as empty functions do not contain semicolon to stop the declarations;

matter123:
A side effect of this is using declarations inside structs have the alias marked as a type.
i.e `using type = T` marks type as `entity.name.type`

matter123:
the following has bad highlighting
```c++
template <class T = value_type>
std::enable_if_t<std::is_trivially_move_constructible_v<T> &&
			std::is_trivially_move_assignable_v<T> &&
			std::is_contiguous_iterator<BidirIt>::value &&
			std::is_contiguous_iterator<OutputIt>::value>
operator()(Allocator &alloc, BidirIt first, BidirIt last,
	   OutputIt dfirst, std::priority_tag<6>) {
		size_t size = std::distance(first, last) * sizeof(value_type);
		std::memmove(addr(dfirst), addr(first), size);
		std::destruct<BidirIt, Allocator,
		              std::destruct_after::move_construct>(alloc, first,
		                                                   dfirst);
}
```
Screenshot from 2019-04-27 22-38-34
(ignore the bold)

matter123:
I think the issue in the last comment can be solved by limiting declarations to be the first thing on a line. I'll try that in the morning.

matter123:
:scope_resolution was moved to the last item in body_includes for struct/class/union.

matter123:
should the `*` or `&` in parameters & declarations such as `const char *argv` be `entity.name.type.modifier` or `variable.other.type_modifier` (i.e. are they part of the type or variable)? Should be answer be different if there is no named parameter (e.g `int strcmp(const char *, const char *);`)?

jeff-hykin:
I'm really excited to look into this PR, I'm going to do it last though since I think it has the largest changes.

jeff-hykin:
Okay, I'm going to work on getting the tests passing for this and get it merged in 👍

matter123:
You might want to hang on, I'm going to submit a PR to fix the sorting of the specs.

jeff-hykin:
> should the `*` or `&` in parameters & declarations such as `const char *argv` be `entity.name.type.modifier` or `variable.other.type_modifier` (i.e. are they part of the type or variable)? Should be answer be different if there is no named parameter (e.g `int strcmp(const char *, const char *);`)?

This is tough, I actually think we should keep it as `keyword.operator` for now and maybe ask alexr about it. It definitely is a modifier, and it would be nice if it was a similar color to the type. I'll get back to you on this.

jeff-hykin:
and okay I'll wait

jeff-hykin:
Screen Shot 2019-05-08 at 10 38 54 PM
Putting this here to keep track of it.

jeff-hykin:
Tomorrow I'll work on trying to get the function pointer syntax working inside function parameters. After that I think it'll just be a matter of updating all the tests.

matter123:
Are you referring to the templated function syntax `template<void(void)`> or something else? `foo(void(*)(void))`?

jeff-hykin:
Just the `function<void(void)>` but hopefully it should work with everything.

matter123:
so a templated function syntax supports parameter names but not default arguments https://godbolt.org/z/24gB-m.

matter123:
code

matter123:
qualified_type might be overly broad in labeling the entire `function<void(void, usertype userval)>` as a type

matter123:
With `function_pointer` again being marked as a parameter. Let's focus again on merging. For most purposes where your function type does not have a name, it is correct, and when it does have a name. The colorization is no worse than it is on the master branch.
Screenshot from 2019-05-09 10-21-30

jeff-hykin:
Okay I agree, I'll try to get it merged in by tonight. Thanks for already getting it working with function  pointer support. In the most recent picture, are the parameter types being matched at? ( the `Event` or the `function` )

matter123:
Sorry I wasn't clear the most recent image 3 hours ago is 1.8.13. The image 14 hours ago is the most recent commit.

jeff-hykin:
I'm thinking after the merge, we can make a range to match the type+variable. The the range can just include the function pointer syntax or any complicated templating/scope resolution.

jeff-hykin:
And ah okay, yeah I'm not too concerned with the named parameter inside the function pointer syntax.

matter123:
The accidental premature merge of Explore-#71 and revert meant that `git merge master` no longer produced a correct merge. This has been rectified. `cpp/generate.rb` now has a successful merge of Explore-#71 and master

matter123:
it looks like issue 11 has a regression `some_var` is marked as `entity.name.type`

matter123:
With regards to `some_var` we can keep existing behavior where a comment is considered a type. Or we can try walking backwards past the comment and fall back to existing behavior if the comment is multi-line. Ruby supports variable length lookbehinds right?

matter123:
Another regression (from an earlier version of this PR)
Screen Shot 2019-05-10 at 09 03 23
The first `compound<typename templated::type *>::type value;` is missing proper highlighting for `typename`, `type`? and `*`. This is not because of the overly broad labeling in qualified_type. The tokens are missing any of their correct scope.

matter123:
I resolved `some_var` by keeping the existing `probably_a_parameter heuristic`. if an identifier follows a block comment it is a parameter.

matter123:
Regression in issue 16.
Screen Shot 2019-05-10 at 11 10 16

matter123:
`qualified type` needs to be aware of multi-word types such as `long double`, `long long int`, `long int` `short int

Added:
```c++
void multi_word_types(signed char, unsigned char, short int, long int, long long,
                      long double, struct a, enum b);
``` to test/fixtures/071.cpp

**Comment by matter123** [comment]
`.zeroOrMoreOf(/unsigned|signed|short|long/.then(@spaces))` turning into `(?:unsigned|signed|short|long\s+)* is a bug right?

**Comment by jeff-hykin** [comment]
Thanks for handling the merging of master. Sadly variable length lookbehinds aren't supported. I think the `compound<typename templated::type *>::type value;` case should probably be changed before merging it in.

**Comment by matter123** [comment]
The tests up to (but not including) issues/055.cpp have been fixed.

**Comment by jeff-hykin** [comment]
templating got messed up a little bit, I'm not sure when this happened though.
function templating messed up

**Comment by matter123** [comment]
I think its because `function_parameters` includes `<>`

**Comment by jeff-hykin** [comment]
Just made a commit for this:
same-line declares issue

**Comment by matter123** [comment]
Ok, that should probably be added to `issues/071.cpp`

**Comment by matter123** [comment]
How did < and > ever get marked as keyword.comparison? None of the template patterns do so.

**Comment by jeff-hykin** [comment]
I think it is from the $base context. In the old grammar all of the `<>`'s were always marked as comparison.

**Comment by matter123** [comment]
so adding `:operators` to `template_call_context` should do it?

**Comment by jeff-hykin** [comment]
Probably, I'm not sure exactly what changed from before though.

**Comment by matter123** [comment]
So I changed in scope resolution. `template_call.without_numbered_capture_groups` to plain `templare_call`

**Comment by jeff-hykin** [comment]
Screen Shot 2019-05-10 at 3 27 29 PM

**Comment by jeff-hykin** [comment]
We might be able to fix that last one by restricting where a function definition can occur. For example, the beinging/root of a file, inside a struct, inside a class etc.

**Comment by matter123** [comment]
So the difference in color is because `characters_in_template_call` was expanded to include `()` to match `<void(void)>`. Removing that makes all four angle brackets white for me.

**Comment by matter123** [comment]
> We might be able to fix that last one by restricting where a function definition can occur. For example, the beinging/root of a file, inside a struct, inside a class etc.

Yeah, It usually is. inside a function, for instance, things are highlighted normally. I'm guessing that lambda bodies are using the root scope.

**Comment by jeff-hykin** [comment]
So it's including `:function_context_c` which is including `:$initial_context`. I'll work on a fix for that

File: cpp/examples.cpp [hunk 1]
```diff
@@ -150,7 +150,6 @@
 //
     auto a = new int(5);
     delete a;
-    new (&a_storage_of_callable.callable) type(forward<Callable>(a_callable));
```
matter123: What makes this invalid?

jeff-hykin:
So that fixed the lambda's one.
Screen Shot 2019-05-10 at 4 08 58 PM

matter123:
Not quite, I reverted the change that added `(` and `)` to characters allowed in a template call. That fixes that but breaks `<void(void)>`

jeff-hykin:
For some reason though the first parameter in lambda captures are different
Screen Shot 2019-05-10 at 4 10 24 PM

matter123:
I think its the `...`

jeff-hykin:
I think qualified type might need to be a range :/

matter123:
it probably should be, but can't

jeff-hykin:
In general yeah, but I think for function parameters it might be able to, and it might be possible for declarations inside root/class/struct/namespace areas too.

matter123:
Declarations is a range, as is function prameters.

cpp/examples.cpp
jeff-hykin: I don't think `new` can be used in that way in the root context.

matter123: Ok, `example.cpp` needs quite a bit of work in that regard then. bare numbers aren't allowed in the root context as well.

matter123:
Screen Shot 2019-05-10 at 14 23 53
Fixed lambda capture.

jeff-hykin:
I think issues like the `compound<typename templated::type *>::type value;` and the `function<void(void)>` could be solved if they used a range-based qualified type instead of a pattern based one.

Templated function calls will probably always need a pattern based qualified type though.

jeff-hykin:
If we can identify the area that a qualified type is supposed to be in then I think we can create a qualified range pattern that will correctly parse any type no matter how complex it is.

jeff-hykin:
Screen Shot 2019-05-10 at 4 15 44 PM

jeff-hykin:
It might be best to split this merge up into smaller changes, and then merge the smaller changes in one at a time.

matter123:
Either skip past the block comment or end the parameter.

Yeah, that's probably a good idea. This is getting pretty complicated.

matter123:
function_pointer is probably the easiest to cut out ill start there.

jeff-hykin:
Sounds Good 👍 
We can probably do
- function pointer
- function definition range (old parameter matching)
- function call
- function definition (new parameter matching)
- variable declarations

jeff-hykin:
I should be free all of today/tonight, I won't make any changes to master so that its easy to merge this in and I'll check periodically to merge pieces of this in. 

If you want to delegate anything let me know, I can probably at least update the tests with each merge.

jeff-hykin:
I think the next thing to work on would probably be to seperate contexts a bit.

Theres a kind a "declaration" context that can't have if statements and other runtime code. I think thats root/class/struct/namespace and maybe some other places. Then there's function body context that can't have function definitions, but can have pretty much everything else. And then there's the parameter context that is just a list of types/parameters.

I think separating these out will help with the declarations. Let me know if you want me to implement them

matter123:
Sure, go ahead and separate the contexts.

jeff-hykin:
This is a lot more difficult than I expected it to be. I'm making good progress, but there are some challenges like the C dependencies. I'm also concerned about objective-c.

Since most everything imports from $initial_context right now, everything kind of needs to split between `declares` (root/class/struct/namespace) and `runtime` (functions, control flow), but it means nothing is inheriting from $inital_context.

All of the C++ changes have been made successfully. I tested, which was fantastic, there's absolutely no way this could've been done without the tests. @matter123  I could use your help with the 017 (universal characters) test, and I had to change 050.cpp since the switch statement needed to be inside of a main function to be valid. I'm not sure how to generate the `050.cpp.json` from the `050.cpp` file. The changes are on the `context_improvements` branch.

I'll conceptually figure out what to do for tonight in regard to the objective-c and C dependence. If there's no good solution, it might be best to just save the context changes for a later time and move forward with the other declaration changes.

matter123:
Yeah, I can work on `issues/017`.

`npm run generate-specs -- issues/050.cpp` will regenerate the spec file for `issues/050.cpp`

matter123:
You actually need a third context.

Scoped function definitions are not valid inside class definitions.
And unscoped constructors are not valid outside class definitions.

See https://godbolt.org/z/eWEns_

matter123:
I'm going to fix 017 on a separate branch because it modifies `entity.name.function` on line 3046 and I want clean tests for that.

jeff-hykin:
Yeah I'm thinking there will be a bunch down the line. An eval context to fix the `(int a)` being invalid, a conditional context for if statements, for `for` loops and while loops, etc. I think I'm mostly done for tonight though.

matter123:
Does
> function definition range (old parameter matching)

Refer to #80?
If so thats blocked by #127 and #163.

jeff-hykin:
Lets close this PR (reopen it if you feel differently) since many of the elements (function pointers, function definitions, qualified_type, type_alias) have been merged in. I think there's a few things we need to cover first like the objective C/C++ issue, and correcting the contexts that we should do before adding the `:declaration` and `:declarations` patterns.

Def. keep the code from this branch though to use it in future PR's.

Issue #80, opened 2019-04-11
Replace legacy function call patterns


matter123:
Many issues such is #79, #76, #51, #50, #43, #26, #18, #17, #2? require changes to the patterns that tag function calls and declarations. These patterns are very long hard to maintain strings. This makes resolving these issue more work than it could be.

jeff-hykin:
Whenever this does get upgraded hopefully we can have it match the general groups of:
`return type area` `name` `parameter area` `qualifier/specifier area` `body area`

matter123:
The following lines are patterns relating to functions that are still strings

line |                  name                        | example match
-----|----------------------------------------------|-------------
1201 | `meta.function.destructor`                   | `Foo::~Foo(`
1224 | `meta.function.destructor.prototype`         | `Foo::~Foo(`
1561 | `meta.function-call`                         | `foo(`
1572 | `meta.function.constructor`                  | `Foo::Foo(`
1598 | `meta.function.constructor.initializer-list` | `Foo::Foo({`
1789 | `c_function_call`                            | `foo(`
2946 | `meta.function`                              | `foo(`
3065 | `entity.name.function`                       | `foo(
3126 | `meta.function.definition.parameters`        | `foo(`

There is still no way of doing back references, correct? If so that blocks converting `1201`, `1224`, `1572`, and `1598`

`1561`, `1789`, `2946`, and `3065` all are similar but have slightly different exclusions, for instance `reinterpret_cast<std::size_t>(__type_name)` matches `1591` to be markes as `meta.function-call` but not the others.

matter123:
@jeff-hykin What is the purpose of `:function_definition`?

jeff-hykin:
I think it's supposed to capture the function name + parameter meta, but I'm not sure if it's actually accomplishing its goal.

It should probably be redone, along with all of the other function stuff.

Hopefully we can shrink everything down to 2 main patterns: function calls and function definitions

matter123:
Function declarations might need to be merged with #71. (see https://godbolt.org/z/lDsYnh). That is probably rare enough that if they are not merged, not too much code would be incorrect.

matter123:
Should special member functions like `Foo::Foo(const Foo &other)` receive any special tagging over normal constructors/operator overloads?

jeff-hykin:
I guess we can tag normal constructors as `.inline` and then `Foo::Foo` as `.out-of-line`? There is probably a better C++ term for out of line things.

[...]

PR #139, opened 2019-05-02
add static_assert

matter123:
This PR adds support for the `static_assert` and `_Static_assert` keywords.

### Why?
`static_assert` is currently (1.8.8) marked as `entity.name.function` when it is rather a function-like keyword, marking it as a keyword fixes the coloring.

Additionally, as #80 progresses this will cause issues if left unattended as static_assert is allowed in places where normal function calls are not.`static_assert(1==1);` is a completely valid C++ program ([goodbolt](https://godbolt.org/z/ox47IF)).
As the distinction is made between function calls and function declarations, `static_assert`'s insides may end up with undesirable coloring.


jeff-hykin:
Something weird is going on with the 
`keyword.static_assert.cpp.c` instead of 
`keyword.static_assert.c`

The grammar class should have automatically fixed that so I'm going to look into what is going on

jeff-hykin:
Took awhile to fix. In the `textmate_tools` the changes were being done as mutations, I changed it to be more of a functional style, all changes are done with a copy, and the copy is returned. I also fixed the other leftover `.cpp` values in the c syntax.

Issue #132, opened 2019-04-26
nested namespaces on same not highlighted correctly


matter123:
Code:
```c++
namespace foo { namespace bar {
	// foo
}} // namespace foo::bar
```
Image:
<img width="323" alt="Screen Shot 2019-04-26 at 14 32 40" src="https://user-images.githubusercontent.com/714007/56837777-75779180-6830-11e9-80e4-67a06f3bd45b.png">

`namespace bar` has the scope `meta.head.namespace.cpp`

jeff-hykin:
<img width="405" alt="Screen Shot 2019-04-26 at 4 41 58 PM" src="https://user-images.githubusercontent.com/17692058/56838032-42d69480-6842-11e9-8b46-1936a4ea6ccb.png">
Looks like scoped namespaces also got messed up.

matter123:
@jeff-hykin This is caused by the commit https://github.com/jeff-hykin/cpp-textmate-grammar/commit/4969c670b2cc0d8c2f680063fe455964d251510a. In the function BlockFinderFor on line 57, the start_pattern for the head range was changed from `/\G/` to `/\G| /`.

This captures a space if present after `{`, causing the lookbehind in the body to fail. Do you know why this change was made?

jeff-hykin:
I know it was made to accommodate  something, I'll try to find it.

matter123:
would maybe `/\G ?/` be correct? that is head optionally captures a space initially.

jeff-hykin:
Yup that seems to work perfect 👌

jeff-hykin:
<img width="315" alt="Screen Shot 2019-05-04 at 4 40 18 PM" src="https://user-images.githubusercontent.com/17692058/57185118-52318f80-6e8b-11e9-8db2-09351a2edc91.png">
Well except for this case

jeff-hykin:
I think thats what the original change was made for.

matter123:
That should probably be a separate release so hopefully, it can still be in Aprils release.

jeff-hykin:
Yeah the switch issue is definitely preferable since it is not a big deal anyways.
It seems like `\G ?` should work if `\G| ` works. Maybe I don't fully understand the behavior of `\G`

matter123:
`:switch_conditional_parentheses` needs to be removed from primary_includes

matter123:
[...]

PR #140, opened 2019-05-03
add suport for enum definitions

matter123:
This PR tags `a` and `b` in the following code snippet.
```c++
enum A {
a, b
};
```
`a` and `b` are tagged as `constant.other.enum` though arguments could be made for other scopes.
Enum values are constant, but it may be advantageous to mark them as variables, as anywhere else they will be treated as such. Additionally `const int some_constant` is still marked as a variable.

Language | Scope
----------|-------
java | `constant.other.enum.java`
c# | `entity.name.variable.enum-member.cs`
typescript | `variable.other.enummember.ts`


jeff-hykin:
Man I wish the new tagging standard was ready. There is no prefect fix for this. Entity isn't great because textmate conventions say "an entity refers to a larger part of the document, for example a chapter, class, function, or tag."
https://macromates.com/manual/en/language_grammars#naming_conventions

I think marking them as a variable is pretty accurate. C++ has the idea of a "constant variable" even though that phrase is a contradiction. I believe the technical definition of a C++ variable is "a named location in memory". And since, I'm pretty sure, the enum is stored in memory, it's accurate.

matter123:
enum variables are adressable but not enum values;
```c++
enum A {
a, b
};
cout << &A::a;
```
does not compile.

jeff-hykin:
Oh that's interesting. Well, in terms of practicality, it would probably be best to mark them as variables since that is probably what they will be marked as in other places of the code.

jeff-hykin:
I guess `variable.other.enummember` is probably the best solution since it is already in use and will fit the most with the other variables in the code.

jeff-hykin:
```
enum A {
a, b
};
```
Adding a space after the `{` breaks the highlighting. This might be a problem with the block finder.

matter123:
Yeah, that is a duplicate of #143 and #132. see https://github.com/jeff-hykin/cpp-textmate-grammar/issues/132#issuecomment-489208354

PR #144, opened 2019-05-05
add bounds to enum

matter123:
See https://github.com/Microsoft/vscode/issues/73292

Screenshot from 2019-05-04 23-34-22


jeff-hykin:
Thanks for catching this

PR #145, opened 2019-05-06
add testing framework

matter123:
This PR adds a testing framework similar to the one used by vscode.

vscode-textmate is used to confirm that fixture files match the spec scopes.

~~This PR does not add the fixtures used by vscode, but conversion should be relatively trivial.~~
vscode's test fixtures are now added

vscode's fixtures https://github.com/Microsoft/vscode/tree/master/extensions/cpp/test


matter123:
Improvements:
- There is now the ability to generate spec files from source fixtures.
- vscode's fixtures should now work without modification.
- a failed test causes `npm test` to fail.
- spec files are now sensitive to leading a trailing whitespace
- zero-width spec objects are supported now

matter123:
Using the vscode tests may not be as trivial as first imagined. Take line 641 of test_c.json https://github.com/Microsoft/vscode/blob/master/extensions/cpp/test/colorize-results/test_c.json#L641
It describes the token "%f%f%f" but that is actually parsed as three seperate tokens.

matter123:
Added a test fixture for most issues. (35 of the 105 total issues were skipped. see `test/fixtures/issues/skipped issues` for why).

The spec files for the test fixtures were generated, as such represents the grammar as it is now (v1.8.9) and now how the open issues should be correctly tagged.

matter123:
If you prefer, I can make spec files in yaml, but they not the most readable thing. They are significantly smaller though (267 KB vs 454 KB).

Example spec
```yaml
- source: int
  scopes:
    - storage.type.primitive.cpp
- source: main
  scopes:
    - entity.name.function.cpp
  scopesBegin:
    - meta.function.definition.parameters.cpp
    - meta.function.definition.parameters.cpp
- source: (
  scopes:
    - punctuation.section.parameters.begin.bracket.round.cpp
- source: )
  scopes:
    - punctuation.section.parameters.end.bracket.round.cpp
  scopesEnd:
    - meta.function.definition.parameters.cpp
    - meta.function.definition.parameters.cpp
- source: '{'
  scopes:
    - punctuation.section.block.begin.bracket.curly.cpp
  scopesBegin:
    - meta.block.cpp
- source: return
  scopes:
    - keyword.control.return.cpp
- source: '1'
  scopes:
    - constant.numeric.decimal.cpp
- source: ;
  scopes:
    - punctuation.terminator.statement.cpp
- source: '}'
  scopes:
    - punctuation.section.block.end.bracket.curly.cpp
  scopesEnd:
    - meta.block.cpp
- source: //
  scopes:
    - punctuation.definition.comment.cpp
  scopesBegin:
    - comment.line.double-slash.cpp
- source: ' a comment /* no nest */'
  scopes: []
```

jeff-hykin:
This is awesome

matter123:
Improved output when a test fails
Screenshot from 2019-05-07 00-27-23

jeff-hykin:
I think this is going to be the best PR for the whole project. The output is perfect, great job on the spacing. This is going to save hundreds of issues in the future. I'll be a lot more comfortable publishing stuff now.

I kind of would prefer YAML if you're okay with it, I'm not sure how much of a change that requires. I think it would be nice because YAML has variables, so whenever we update a scope name, we can update it in one place in the YAML.

Also, now that we have an `npm test` I thought it'd be best to also create a debugging config. So now to test, we can just startup the built-in VS Code debugger.

matter123:
It's not a lot of work, ill push to this branch soon.

PR #153, opened 2019-05-09
Fix/#150

matter123:
the check for `//` and `/*` was not a lookahead, this caused the starting token to be consumed and the comment not treated as such


jeff-hykin:
I changed the way enums are matched, but I realized it's not a good solution so I'm going to merge this fix to overwrite it.

PR #155, opened 2019-05-09
Fix/sort spec keys

matter123:
the keys in spec files were supposed to be sorted `source`, `scopesBegin`, `scopes`, `scopesEnd`


matter123:
Sorry I forget I was on Explore-71 you should probably cherry-pick

jeff-hykin:
I can't believe I didn't notice that. I'll revert the merge and then cherry pick

Issue #30, opened 2019-03-16
meta.processor and meta.struct are never closed


jeff-hykin:
https://github.com/atom/language-c/issues/267
```
#pragma once

#include <type_traits>

#define IsPointDef(...) \
    template<> \
    struct IsPoint<__VA_ARGS__> \
        {\
        static const bool isPoint = true;\
                }

#define ArrayBasedPointDef(T) \
    IsPointDef(T); \
    template<> \
    struct IsArrayBasedPoint<T>:public std::true_type \
        {};



#define XYBasedPointDef(T) \
    IsPointDef(T); \
    template<> \
    struct IsXYBasedPoint<T>:public std::true_type \
{};

#define TypeTAndUIsPoint \
    template<typename T, typename U, class = typename std::enable_if<IsPoint<T>::isPoint>::type, class = typename std::enable_if<IsPoint<U>::isPoint>::type>

namespace Navigation
{
    namespace Utils
    {
        template<typename T>
        struct IsPoint
        {
            static const bool isPoint = false;
        };

        template<typename T>
        struct IsArrayBasedPoint
        {
            static const bool value = false;
        };

        template<typename T>
        struct IsXYBasedPoint
        {
            static const bool value = false;
        };

    }
}
```

matter123:
Probably blocked by https://github.com/Microsoft/vscode/issues/20488
There needs to be a way of processing multiline macro that terminates when /\\$/ no longer matches. Begin-end rules can have the nested include "overshoot" past the end, that is what is causing this.

matter123:
The following can parse a multi-line macro-like structure
Unfortunately, it prevents using begin-end rules that cross multiple line

```json
		{
			"begin": "#define",
			"end": "(?<!\\\\)\\n",
			"name": "bold.debug",
			"contentName": "debug.debug",
			"patterns": [
				{
					"name": "single_line.debug",
					"match": "(?:\\G|^)[^\\n]+(?=\\\\$)(?<!\\\\)"
				},
				{
					"name": "single_line.debug",
					"match": "(?:\\G|^)[^\\n]+(?=$)"
				}
			]
		},
```

jeff-hykin:
closing in favor of #136

PR #159, opened 2019-05-10
Fix #26

### LINKED ISSUE #26: constructor gets confused by function pointers

matter123:
This is the Fix for #26 with updated test extracted from #135.


jeff-hykin:
Screen Shot 2019-05-10 at 5 45 43 PM

jeff-hykin:
I'm not sure why the `}` is missing highlighting

matter123:
the a isn't valid c++ code

matter123:
https://godbolt.org/z/Jwinbc

jeff-hykin:
For the brackets at the end of the function pointer, would that only be for declaring an array of function pointers?

matter123:
No its actually for declaring the the return type is an array https://cdecl.org/?q=int+%28*foo%29%28%29%5B%5D

jeff-hykin:
haha that is a great link

matter123:
Oh sorry the current_array brackets yes declare an array of function pointers.

jeff-hykin:
Do you understand this? `int (*(*foo)(void ))[3]`
https://cdecl.org/?q=int+%28*foo%29%28%29%5B%5D
I'm still not sure what exactly is being declared. I mostly dont understand the `[3]`

matter123:
So it helps to remember that its the same as `int *(*foo)(void )[3]` which is a pointer to a function take takes nothing and returns a pointer to an array of 3 integers.  which is not actually valid https://godbolt.org/z/8sxwSh

matter123:
No idea why cdecl's example is not actually a valid declaration.

matter123:
There should be support for `int *(*foo)()` and `int &(*foo)()` right?

jeff-hykin:
yeah, both of them compile

matter123:
Ok added support and update test

matter123:
Screenshot from 2019-05-10 16-54-42

jeff-hykin:
So is something like `int (*fcnPtr[])(void);` an array of function pointers?

matter123:
Yep

matter123:
Your dumb ptr/ref allows for `int **&(*foo)();` which isnt valid

jeff-hykin:
👍 working on a fix

matter123:
Also, it looks like the * is no longer tagged.

jeff-hykin:
Not sure which of these is valid according to the spec
Screen Shot 2019-05-10 at 7 15 00 PM

jeff-hykin:
Ah I just realized I left the `[]` in those last two examples

matter123:
yeah because rr value references aren't a thing.
so any number of `*` then up to 2 `&`

matter123:
It should just be
```
optional_type_modifier = maybe(@spaces).then(
            match: zeroOrMoreOf(/\*/).maybe(/&&?/),
            # TODO maybe change see https://github.com/jeff-hykin/cpp-textmate-grammar/pull/135#issuecomment-490727262
            tag_as: "keyword.operator",
        ).maybe(@spaces)
```

Edit: yours is betetr i didn't see it yet.

cpp/generate.rb
```diff
@@ -28,6 +28,16 @@
             match: /,/,
             tag_as: "comma punctuation.separator.delimiter"
         )
+    # TODO maybe change this tag see https://github.com/jeff-hykin/cpp-textmate-grammar/pull/135#issuecomment-490727262
+    # TODO eventually move this outside of the # Utils section
+    ref_deref_definition_pattern = newPattern(
+            match: zeroOrMoreOf(/\*/.maybe(@spaces)),
+            tag_as: "keyword.operator.dereference"
```
matter123: that's a pointer, not a dereference.

jeff-hykin: True. Okay, I guess if were going to make it accurate we should decide on something other than `keyword.operator`. Do you think `storage.modifier.pointer` and `storage.modifier.reference` are accurate?

matter123:
LGTM. I'll update the tests.

matter123:
Is it intentional that the * in `*fncPtr` only has a punctuation scope?

matter123:
This broke tests:
-  `issues/030.cpp`
- `issues/088.cpp`

jeff-hykin:
> Is it intentional that the * in `*fncPtr` only has a punctuation scope?

It was, but its up for debate. I think punctuation (in contrast to operator) is for anything thats telling the compiler where/what something is. Its debatable that it could be part of the `storage.modifier` since its a function pointer.

matter123:
any idea why the test for #30 is broken?

jeff-hykin:
Looking into it now, I'm not immediately sure why though

jeff-hykin:
Oh I forgot about `# TODO this is here because it needs to activate inside of function-pointer parameters. Once function-pointer syntax is implemented, remove it from here` and I think that actually might be the cause

matter123:
Ok, im fixing up test 26.

matter123:
Screenshot from 2019-05-10 18-14-11

matter123:
Anything else that needs to happen before his gets merged in?

jeff-hykin:
I think everything is good 👍

PR #161, opened 2019-05-11
Add type alias

matter123:
Adds a `type_alias` pattern and tests. Extracted from #135.


jeff-hykin:
If you can add some unit tests to `:type_alias`, I'll go ahead and merge it in.

I still don't fully understand the `:qualified_type` pattern but I'll figure it out later

matter123:
Sure.

matter123:
The unit tests aren't quite sufficient yet

matter123:
So `:qualified_type` is supposed to match a fully storage, namespace, and template qualified type.
that is `struct A::B::C::D` or `std::enable_if<std::numeric_limit<long long int>::digits > 10, bool>` It doesn't match type qualifiers like const or static or restrict, or type modifiers like `**` and `&` so that it can be used in more places. The hope/plan is that much in the same way `variable_name` is used anywhere a name is expected, `qualified_type` is used whenever a type is expected.

PR #164, opened 2019-05-11
initial patterns for trigraph support

matter123:
This adds to `textmate_tools.rb` helper patterns to support the 10 characters that have digraph and trigraph alternate token representations.

The following tokens have an alternate representation

token | digraph | trigraph 
-------|------------|-----------
\## | %:%: |??=??= | 
\# | %: | ??= |
[ | <: | ??(
] | :> | ??)
{ | <% | ??<
} | %> | ??>
\ || ??/
^||??'|
\|||??!|
~||??-|

~~Currently no support for trigraphs have been added to the patterns~~
blockFinderFor now uses the alternate operator patterns

default scope names have been removed as they cause issue with lookBehindFor, even when using `without_numbered_capture_groups`.


jeff-hykin:
I think its best to keep the textmate_tools.rb language agnostic so I moved the trigraph stuff to an the shared folder

PR #169, opened 2019-05-15
Fix access specifier scope has colon

matter123:
in the example
```c++
class Foo {
  public:
  /* snip */
};
```
`public` had the scope `storage.type.modifier.access.control.public:`.
even worse when there is a space between `public` and `:` `public` had the scope `storage.type.modifier.access.control.public.cpp :.cpp`.

This PR prevents the `:` from ending up as part of the scope name.


matter123:
Closing as the branch was dirty. See #170.

PR #178, opened 2019-05-20
Add regex pattern coverage

matter123:
This PR adds a pattern coverage report to the end of npm test. Example output:
Screenshot from 2019-05-19 22-54-30

This was achieved by wrapping the requested OnigLib with a object that returns a custom implementation of an OnigScanner. This scanner manually tests for the matching pattern. By doing this the custom scanner can record what patterns were chosen and what patterns failed to match at all. To ensure that the test quality is not compromised. A genuine OnigScanner is used to give the results to vscode-textmate. A comment containing the original location of each pattern is added in by rewriteGrammar to different patterns that are identical but in different locations,

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -278,6 +278,12 @@
                 "esprima": "^4.0.0"
             }
         },
+        "json-source-map": {
+            "version": "0.4.0",
+            "resolved": "https://registry.npmjs.org/json-source-map/-/json-source-map-0.4.0.tgz",
+            "integrity": "sha1-7qg3/jzi8r/VsTaHd5QGNUQjw1U=",
+            "dev": true
+        },
         "json-stable-stringify": {
             "version": "1.0.1",
             "resolved": "https://registry.npmjs.org/json-stable-stringify/-/json-stable-stringify-1.0.1.tgz",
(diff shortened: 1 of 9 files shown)


jeff-hykin:
Wow that's really impressive. I don't quite understand how what it means by match failure though. What exactly would need to be done to get 100%?

matter123:
the match failure percent is what percent of patterns that were tested and failed at least once. So if a pattern never failed to match then it would prevent the percentage from being 100%

matter123:
You'll want to modify getRegistry to not inject on line 10 of test if coverage is not passed. collecting the data is kinda slow.

matter123:
Basically, to get to 100% you need:
for match failure: for every pattern to potentially be considered but fail to match at least once 
for match not chosen: for every pattern to be the only pattern considered or match but not be chosen at least once
for match chosen: for every pattern to be considered and then chosen at least once.

jeff-hykin:
On one of my last teams we were required to get 90% coverage and to do that we would actually make the code worse because some of the code was incredibly hard to test and it was easier to remove functionality than to write a test for it. If I'm being realistic, that probably not a concern for this project, but in general I'm a bit hesitant towards coverage metrics.

So long as we don't measure success by coverage % I think it'll be a useful tool.

matter123:
Yeah, I definitely agree that we shouldn't remove features to get the coverage % up. It's more of something to keep track of, as there can be potential regressions and bugs hiding in the uncovered portion.

Issue #180, opened 2019-05-20
Object members highlighting


alekseyt:
This is a question related to vscode 1.34.

I'm trying to highlight class/struct members like this:

![image](https://user-images.githubusercontent.com/993454/58036505-f0804f00-7b33-11e9-8f57-f5e19ca2a3b5.png)

I'm tried to highlight `variable.other.object` and this kind of works, but it highlights part i didn't want to highlight:

![image](https://user-images.githubusercontent.com/993454/58036635-3dfcbc00-7b34-11e9-93cc-874ac68defb7.png)

Is there a way to separately highlight object and its members?

Thanks.

matter123:
See #59 in particular https://github.com/jeff-hykin/cpp-textmate-grammar/issues/59#issuecomment-481918973

jeff-hykin:
Here's what you can do. 
`variable.other.object.access` // should tags the first word
 `variable.other.object.property` // should tag the middle words
`variable.other.property` // should tag the last word

If that isn't the case (I'll check at home) then something needs to be fixed.

alekseyt:
Thank you, i had moderate success with `variable.other.object.property`:

![image](https://user-images.githubusercontent.com/993454/58042384-3bee2980-7b43-11e9-8a28-5d056b5d5250.png)

It does that, but surprisingly it highlights `z->` entirely (or `z.` if that matters). I've read comments in #59 and tried to override color for `punctuation.separator.pointer-access.cpp` or even `punctuation.separator`, but it doesn't seem to help. I suspect maybe `->` is being captured with `z` or something, not sure about this.

Maybe this is already fixed, apparently syntax file was already updated in vscode's master and revision in 1.34 might be somewhat old. I replaced it from master and it fixed some other issues i've seen in 1.34, but not this one. But `variable.other.object.property` definitely seems like a way to go.

matter123:
That's #130. I'll try to get a PR for it in the next 12 hours.

alekseyt:
I'm closing this as a duplicate then. Thank you very much.

PR #182, opened 2019-05-20
Fix for #130 by duplicating partial_member

### LINKED ISSUE #130: subsequent member access tagged a variable;
Code:
```c++
a->b->c = 1;
a.b.c = 1;
```
Image:
Screen Shot 2019-04-26 at 1 11 48 PM

`->` and `.` member access operators after the first one are part of the same token as the member that precedes it.

matter123:
This Fixes #130 and #180. 
Partial member had to be duplicated in order to tag just the portion of partial_member that was supposed to be tagged.


alekseyt:
I checked this locally and this fixes highlighting in member access chains like `x->y->z` or `ptr->z->y.x->w`.

PR #184, opened 2019-05-21
Fix #183

### LINKED ISSUE #183: namespace is missing meta.head.namespace scope
See #179. the head of a namespace block does not have the meta.head.namespace scope.

matter123:
This adds namespace aliases and moves the namespace name into the namespace head pattern.

### Namespace Alias:
this tags `namespace name = qualified-namespace ;` as `meta.declaration.namespace.alias` with appropriate scopes for each component.

### Namespace blockFinderFor changes:
with namespace aliases, the only meaning on `namespace` now is the start of a namespace block. This means that the start pattern of the namespace blockfinder for can be simplified to just a bound /namespace/. The tagging of namespaces was moved into the head includes.

@alekseyt this should allow you to target `scoped` in `namespace scoped::namespace {}` with `meta.head.namespace.cpp entity.name.type.namespace.scope-resolution.cpp`


alekseyt:
Thank you. Can confirm C++17 nested namespaces can now be colored uniformly using `meta.head.namespace.cpp` to override color of `scope-resolution` in the beginning (head) of namespace, namespace aliases are highlighted too.

jeff-hykin:
Well I was just going to add support for localized scope resolution names. I did that, but then the tests were broken for function calls, and then for function definitions. So I updated function calls and definitions (got rid of the legacy versions), then there was a minor problem with templated functions so I tweaked the `template_call` pattern.

A lot of tag names changed, and in general a lot of core patterns changed so this commit definitely could use your review. Sorry that it's one giant commit.

The change could still use cleaning up, I think there are still some legacy function patterns hanging around.

cpp/generate.rb
```diff
@@ -55,6 +55,29 @@
             tag_as: "storage.modifier.reference"
         ).maybe(@spaces)
     )
+    functionCallTemplate = ->(repository_name:nil, match_name: nil, tag_name_as: nil, tag_content_as: nil, tag_parenthese_as: nil) do
```
matter123: this should be renamed to not include the word template. maybe `generateFunctionCall` ?

jeff-hykin:
One known regression is the type-casting operators are now marked as templated function calls.

cpp/generate.rb
```diff
@@ -480,8 +530,10 @@ def blockFinderFor( name:"", tag_as:"", start_pattern:nil, needs_semicolon: true
     # because it's embedded inside of other patterns
     cpp_grammar[:template_call_innards] = template_call = newPattern(
         tag_as: 'meta.template.call',
-        match: lookBehindToAvoid(/</).then(/</).lookAheadToAvoid(/</).zeroOrMoreOf(characters_in_template_call).then(/>/).maybe(@spaces),
-        includes: [:template_call_context]
+        # if we add readable-support for recursive regex, then we might be able to use /<((?>[^<>]+|(\g<#groupNumberOfThisGroup>))*)>/ 
+        # to match the characters in the middle of a template call
```
matter123: We also would need to confirm that vscode-textmate supports recursive regex.

cpp/generate.rb
```diff
@@ -593,28 +645,51 @@ def blockFinderFor( name:"", tag_as:"", start_pattern:nil, needs_semicolon: true
...
-        tag_as: "meta.scope-resolution",
-        match: preceding_scopes.then(
-                match: variable_name_without_bounds,
-                tag_as: "entity.name.type.namespace.scope-resolution"
-            ).maybe(@spaces).maybe(
-                template_call
-            ).then(
-                match: /::/,
-                tag_as: "punctuation.separator.namespace.access"
+    scopeResolutionFinder = ->(tag_extension, grammar_name, hidden_grammar_name) do
```
matter123: It seems that hidden_grammar_name should just be computed form grammar_name.

cpp/generate.rb
```diff
@@ -734,12 +792,42 @@ def blockFinderFor( name:"", tag_as:"", start_pattern:nil, needs_semicolon: true
...
+                cpp_grammar[:scope_resolution_function_definition]
+            ).then(
+                avoid_invalid_function_names
+            ).then(
+                match: variable_name_without_bounds,
+                tag_as: "entity.name.function.definition"
+            ).maybe(@spaces).lookAheadFor(/\(/)
+        ),
+        needs_semicolon: false,
+        primary_includes: [],
```
matter123: repeating `primary_includes: []` is just noise, same for `secondary_includes`

matter123:
> One known regression is the type-casting operators are now marked as templated function calls.

I cant repo
Screenshot from 2019-05-21 20-19-51

cpp/generate.rb
jeff-hykin: yeah, and we might actually want to rename the `Finders` to `Generate` since they're kind of in the same space.

cpp/generate.rb
jeff-hykin: yeah, and I'd only give it a 50% chance that it does.

cpp/generate.rb
jeff-hykin: Okay, I did that at first, it just makes it a little weird when you are trying to search for one of the grammar-repo names because it doesn't actually exist anywhere.

The code I had before just appended `_innner` to the name. If you don't think thats confusing I'll add it back

cpp/generate.rb
jeff-hykin: Ah thanks, I pasted all the inputs to fill them out and I forgot to clean it up afterwords.

jeff-hykin:
Oh I guess one of the last changes I made fixed the type casting ¯\_(ツ)_/¯

cpp/generate.rb
matter123: maybe `_inner_generated` to make it a bit more clear that its a synthetic tag name.

PR #192, opened 2019-05-22
move primray_includes to after head_includes

matter123:
`head_includes` `start_pattern` requires that nothing else matches before it.


jeff-hykin:
This was actually done intentionally so that, if there's an edge case where we do need to override head, we can do it without changing the `generateBlockFinder` function.

Is there a reason we need an additional context that comes before body but after head?

matter123:
No, if primary_includes is intended to allow for replacing the head then its fine.

PR #197, opened 2019-05-23
Adds new regex patterns

matter123:
This implements possessive/reluctant modifiers, atomic groups, and repetition as described in https://github.com/jeff-hykin/cpp-textmate-grammar/issues/127#issuecomment-493330684 and https://github.com/jeff-hykin/cpp-textmate-grammar/issues/127#issuecomment-493337504

Example usage:
```ruby
.then(
            match: /&/.maybe(@spaces),
            atLeast: 0.times,
            atMost: 2.times,
            tag_as: "storage.modifier.reference"
        )
```


jeff-hykin:
This is great. It's going to take me a bit to get to it.

jeff-hykin:
If you can change the names to snake case like `at_least`, and if you're confident there aren't bugs I'll merge it in without looking at it so you can use it on other issues.

Then when I can I'll just come back and look at it later.

matter123:
The addition of support for the new regex patterns didn't break any existing patterns and i am reasonably confident the new patterns are working as intended.

matter123:
Added support for the implied requirement that if `how_many_times?` is an Integer or Enumerator than it should match that many times.

Issue #125, opened 2019-04-24
have grammar check at the end to make sure that all of the included repository_names are actually valid repo names


matter123:
Making a separate issue to track this.

This TODO would have prevented #123.

jeff-hykin:
Pushed an initial fix for this on master. But this only works for :$initial_context.

Its going to be a lot more work to check the names of all the pattern's/ranges #includes because we have to wait until the grammar is being exported before checking.

PR #202, opened 2019-05-25
add linter

matter123:
This adds a simple 2 pass linter over the produced grammar.
The linter is given an entire textmate rule and is allowed to keep context.
This is used in this commit to check for unresolved and unused names.

Linters could be added to check scope names for spelling mistakes, or consistency.
There could be a linter to ensure `null` is never a key as that breaks vscode-textmate.

Fixes: #125

Sample Output:
Screenshot from 2019-05-25 00-57-02

### DIFF
--- a/lint/index.js
+++ b/lint/index.js
@@ -17,7 +17,10 @@ const fs = require("fs");
  * }} TextMateRule
  */
 
-const linters = [require("./linters/unresolved")];
+const linters = [
+    require("./linters/unresolved"),
+    require("./linters/spell-check")
+];
 
 const grammar = JSON.parse(fs.readFileSync(process.argv[2]).toString());
 /**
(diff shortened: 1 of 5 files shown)


matter123:
Added a spell checking linter.
Screenshot from 2019-05-26 13-33-54

The linter takes `name` and `contentName` from each rule. Splits by [" ", ".", "-"], ignores "$N" and removes "like" from the end of each part. Then, checks each part to ensure that it is spelled correctly.

Has currently detected that  `inhertance` and `specificer` are spelled incorrectly.

jeff-hykin:
> `null` is never a key as that breaks vscode-textmate.

😆wow 

> Has currently detected that inhertance and specificer are spelled incorrectly.

😬 oops

Issue #193, opened 2019-05-22
function definition pattern matches inside function definition scope


matter123:
Example:
```c++
foo bar() {}
struct foo() {
	foo bar() {
		foo bar();
	}
};
```
![Screenshot from 2019-05-22 13-56-37](https://user-images.githubusercontent.com/714007/58208442-95686c80-7c99-11e9-8d6c-59147a85d92e.png)
The bar on line 4 cannot be a function definition as functions are not allowed to be nested. bar is rather a default constructed foo object.

jeff-hykin:
Thats really strange. The function definition pattern includes the `:function_call` before the `:$initial_context` so it should give it higher priority. For some reason it doesn't though (removing the $initial_context allows the function call to match)

matter123:
it's the best match rule. with function definition now matching the type. the function definition matches earlier in the line than the function call witch matches at `bar` See https://github.com/jeff-hykin/cpp-textmate-grammar/blob/master/test/src/pattern-coverage/onigScanner.js#L49

jeff-hykin:
Well I guess now is as good a time as ever to create a pattern for a default constructed object.

jeff-hykin:
That or we could try to figure out a $initial_context solution. The problem being that we can't just use a different scope entirely because it will probably break other syntaxes like Objective C++

matter123:
This is a relatively minor issue as per #36 bar would have already been marked as a function. This should resolve itself as #81, #138, and contexts are resolved.

matter123:
`foo * bar()` where foo is a variable and bar is a function call causes `foo *` to be as a type that is a pointer to foo. See #204

Issue #204, opened 2019-05-25
"<variable> * <function>" highlighted incorrectly


WEGFan:
VS Code version: 1.34.0
Extension version: 1.9.0 ~ 1.9.5

Sample code
```cpp
int fun() 
{
    return 1;
}

int main()
{
    int a, b; 
    a = b * fun();
    return 0;
}
```
![Snipaste_2019-05-26_01-01-39](https://user-images.githubusercontent.com/7558201/58372570-db4f5a00-7f51-11e9-8f38-400cac577ac8.png)

matter123:
Can you describe what you expect the highlighting to be?

WEGFan:
"b" and "*" should not be highlighted, it seems that they were treated as a type and a pointer.
![image](https://user-images.githubusercontent.com/7558201/58372961-66335300-7f58-11e9-9335-48cb375c2384.png)
![image](https://user-images.githubusercontent.com/7558201/58372966-84994e80-7f58-11e9-9b22-05dc3d05d380.png)

matter123:
Duplicate of #193. I'll try to get a fix in today.

PR #205, opened 2019-05-25
remove function definition from fiunction_definition

matter123:
Fixes: #193, #204

Screenshot from 2019-05-25 11-15-00


jeff-hykin:
Thanks for getting this fix in.

I think this is the right move (it's fixing the root problem; the context) so I'm planning on merging it in. But as a heads up, this is probably going to break the Objective-C++ syntax because it's not using $initial_context.

So creating a fix for objective C++ will probably become higher priority after this.

PR #214, opened 2019-05-28
New folder structure to accommodate Objective-C  

jeff-hykin:
@matter123 I won't merge this in yet because it would break the other merge requests, but I wanted to get your opinion on this new structure.


scripts/generate.rb
```diff
@@ -0,0 +1,6 @@
+Dir.chdir __dir__ # go to this directory
+Dir.chdir ".." # go up one dir__
+
+for each_dir in Dir["source/languages/**"]
+    Process.wait(Process.spawn("ruby", each_dir+"/generate.rb"))
```
matter123: this should exit with status code 1 if the subprocess exits with status code 1

matter123:
`test/src/` should become `test/source/` to match.

source/languages/c/generate.rb
```diff
@@ -1,6 +1,6 @@
-require_relative '../textmate_tools.rb'
+require_relative '../../textmate_tools.rb'
 require_relative './tokens.rb'
-require_relative '../shared/numeric.rb'
+require_relative '../../shared_patterns/numeric.rb'
```
matter123: with scripts/generate.rb now, would it be better to add /source/ to the load_path and use?:
`require textmate_tools.rb`
`require shared_patterns/numeric.rb`

jeff-hykin: I didn't even know `load_path` existed, thats really cool. I think that'd be a much better solution

jeff-hykin:
> The `tags.txt` files are a generated file. Would it make more sense to move them out of `/source/`?

I was thinking the same thing, I guess we can just have a tags folder and then point them there

matter123:
It looks like `entity.name.type.namespace` came back.

source/languages/c/generate.rb
jeff-hykin: Well now that I've implemented it, it seems decently harder to understand whats going on
```
$LOAD_PATH << File.join(File.dirname(__FILE__), "../../") # add source to the load path 
require 'textmate_tools.rb'
require 'repo_specific_helpers.rb'
require 'shared_patterns/numeric.rb'
require 'shared_patterns/predefined_macros.rb'
require_relative 'tokens'
```

Do you think this is any better?
```
source_dir = "../../"
require_relative source_dir + 'textmate_tools.rb'
require_relative source_dir + 'repo_specific_helpers.rb'
require_relative source_dir + 'shared_patterns/numeric.rb'
require_relative source_dir + 'shared_patterns/predefined_macros.rb'
require_relative './tokens.rb'
```

jeff-hykin:
> entity.name.type.names

I think that might be from merging the predefined macros branch, I'll take a look

source/languages/c/generate.rb
matter123: Yeah, the second version is better.

jeff-hykin:
Nope it was my mistake on restoring the `using` inside of the attributes

scripts/generate.rb
```diff
@@ -3,7 +3,9 @@
 
 for each_dir in Dir["source/languages/**"]
     exit_code = Process.wait(Process.spawn("ruby", each_dir+"/generate.rb"))
-    if exit_code != 0
```
matter123: Why was this changed?

jeff-hykin: Actually I could use some help on this section. 
There's this really strange bug. When using backticks in ruby or running the command in the command line, I get exit code `0` for the `cpp/generate.rb`.

However, the same command using `system()` or `Process.spawn()` will return a `76XXX` error code like `76101` and every time I run it the error code changes even though the process finishes successfully.

matter123: `Process.wait` returns the process id of the waited on process. use `$?` to get the exit code.

jeff-hykin: Haha thanks, I should've caught that.

package.json
```diff
@@ -18,6 +18,17 @@
...
+        "C++",
+        "cpp",
+        "c",
+        "objective-c",
+        "objective-cpp",
+        "objective-c++",
+        "syntax",
+        "textmate",
+        "highlighting"
+    ],
```
matter123: maybe added color(ing), a couple of other syntax extensions have that (tree sitter, c++ reloaded, yml, syntax highlighter)

source/repo_specific_helpers.rb
```diff
@@ -0,0 +1,7 @@
+def saveGrammar(grammar)
```
matter123: shoulden't this be part of textmate_tools?

jeff-hykin: Ideally textmate_tools.rb will become library/ruby gem and mapping it to a specific folder structure would kind of break that.

package.json
jeff-hykin: done 👍

package.json
```diff
@@ -1,11 +1,11 @@
 {
     "name": "better-cpp-syntax",
-    "displayName": "Better C++ Syntax",
-    "description": "The bleeding edge of the VS Code C++ syntax",
+    "displayName": "Better Syntax",
+    "description": "The bleeding edge of some VS Code syntaxes",
```
matter123: maybe list the languages or say C-like. "some VS Code syntaxes" is pretty vague.

jeff-hykin:
Thanks for doing that switch to snake case 👍

source/languages/c/generate.rb
```diff
@@ -1,19 +1,23 @@
...
 
 c_grammar = Grammar.new(
+    name: "C",
+    scope_name: "source.c",
+    version: "",
```
matter123: Is no version information okay?

jeff-hykin: I think so. I should actually probably make it optional in the `textmate_tools.rb`. VS Code overwrites the version information whenever they pull it in.

matter123:
Anything else you think should be in this merge?

jeff-hykin:
Yeah, I'm looking at Objective-C++ examples and it still looks pretty broken :/

matter123:
So, how much is sharable between C++ and Objective-C++?

jeff-hykin:
Apparently a lot https://stackoverflow.com/a/4672776
Sadly the Apple link he provided is broken

Issue #222, opened 2019-06-04
Whitespace before #define causes #define to get wrong scope


alexr00:
![image](https://user-images.githubusercontent.com/38270282/58884253-4f22fc80-86e0-11e9-8f1e-ebddb969ab78.png)

PR #223, opened 2019-06-04
make source_wrapper never consume characters

matter123:
When the source wrapper pattern consumes a character it makes patterns that match using the start anchor no longer work when that pattern is the first line. This PR modifies source_pattern to not consume any characters and provides an end pattern that is not possible to match. These two changes combined, ensure that the source wrapper matches every part of the document.

Fixes: #222


jeff-hykin:
Thats a creative solution 👍, much better than the previous source_context

PR #227, opened 2019-06-07
Add assembly support.

matter123:
This PR adds support for marking embedded assembly as such.

Example:
Screenshot from 2019-06-06 17-15-26

This relies upon another grammar to provide the actual syntax highlighting of the embedded language.

This is ~~not~~ currently suitable for merging because:
 - ~~when no grammar provides syntax highlighting it fails to fall back to marking the strings~~

### DIFF
--- a/source/shared_patterns/assembly.rb
+++ b/source/shared_patterns/assembly.rb
@@ -1,0 +1,46 @@
+def assembly_pattern()
+    return PatternRange.new(
+        start_pattern: newPattern(
+            match: /\b(?:__asm__|asm)\b/,
+            tag_as: "storage.type.asm",
+        ).maybe(@spaces).maybe(
+            match: /volatile/,
+            tag_as: "storage.modifier"
+        ).maybe(@spaces).then(
+            match: /\(/,
+            tag_as: "punctuation.section.parens.begin.bracket.round.assembly",
+        ),
+        end_pattern: newPattern(
+            match: /\)/,
+            tag_as: "punctuation.section.parens.end.bracket.round.assembly"
+        ),
+        includes: [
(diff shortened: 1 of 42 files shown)


source/shared_patterns/assembly.rb
```diff
@@ -0,0 +1,50 @@
...
+                tag_as: "string.quoted.double",
+                tag_content_as: "meta.embedded.assembly",
+                includes: [
+                    "source.asm",
+                    "source.x86",
+                    "source.x86_64",
+                    "source.arm",
+                    # this is here because this pattern will not match if all of the includes above are not found.
+                    # it never matches to ensure no overshoot
+                    lookAheadFor(/not/).then(/possible/)
```
jeff-hykin: So what would happen if this was removed?

matter123: If that pattern is removed, and none of the includeded grammars exist, then the entire string is one token with no scopes.

matter123: See the spec files in the original commit for what that looks like.

matter123: Rereading the comment, I realized it is a bit unclear. If the includes consists entirely of includes that cannot be resolved the outer pattern tags will not be applied. If the dummy pattern is removed the entire string gets no special tagging at all.

jeff-hykin:
can the `asm()` take a variable? rather than a string

matter123:
Nope [[dcl.asm]/1](http://eel.is/c++draft/dcl.asm#1) [Gcc](https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html)

PR #234, opened 2019-06-09
WIP: Add C style cast support

matter123:
This PR marks a possible type between parentheses as a type cast.

More testing is needed before merging.


matter123:
The latest commit make `return (some_var);` and `new (some_type);` not type caste by preventing the type caste from being followed by a semicolon.

jeff-hykin:
I updated this with master, feel free to revert it if I missed changes somewhere, I just didn't want to leave this branch behind.

PR #253, opened 2019-06-23
make space after = in type alias optional

matter123:
The unknown type expression pattern required at least one character before it matched, this caused issues when a type alias was formatted such that the type expression was on a separate line.

This PR makes the unknown type expression able to match even with no characters.


jeff-hykin:
I manually merged this in since it was a one character change

PR #277, opened 2019-07-02
select test based on modified files

matter123:
This PR modifies the default behavior of `npm test`. With this PR, when you have uncommitted changes to a grammar file, only the tests that would be affected by the change are ran. If there are no uncommitted changes to grammar files, the behavior of `npm test` is not modified. In particular, if `git status` shows a modified `.tmLanguage.json` file, than only the fixtures whose file extension matches the grammar's `fileTypes` are ran.


jeff-hykin:
I think this is great. Sometimes I need to commit WIP changes though, could you add a `testall` command that had the original functionality?

matter123:
the option `--generate-all` has been renamed to `--all` `npm test -- --all` performs default filtering on the tests. `npm run testall` is an alias for `npm test -- --all`

PR #294, opened 2019-07-08
WIP: Add script to generate duplicate grammars for embedding

matter123:
`scripts/duplicateForEmbedding.js` adds a generic way to support any number of early bailout patterns.


matter123:
Closing in favor of #295

PR #295, opened 2019-07-08
Add script to generate duplicate grammars for embedding

matter123:
This PR is from the updated master.

`scripts/duplicateForEmbedding.js` adds a generic way to support any number of early bailout patterns.

~The grammar has been modified to use the macro embedded grammar for macros.~


jeff-hykin:
This looks nice 👍 I'll work on getting it merged in in the next few days

matter123:
The Extension Development Host scopes don't match what the tests say the scopes are. I'm guessing that I'm not announcing the syntax file correctly.

matter123:
While this, might work and is probably still useful if another grammar needs to embed c++. This is not the best route to go with macros.

There are several disadvantages:
 - This duplicates non-range patterns
 - If an embeddable grammar is made it will need its own macro-fied version

matter123:
If you prefer I can rewrite this in ruby, but otherwise, I believe it works well enough to be merged in.

jeff-hykin:
> This looks nice 👍 I'll work on getting it merged in in the next few days

Oops 😬

jeff-hykin:
I'll merge this in before I do anything else, but yeah in the future lets turn it into feature inside the ruby gem. I'm fine with stuff being done in javascript in general though, its basically the universal programming language.

I was imagining the bailed-out patterns as part of the language-syntax file, so you could take any subset language, for example regex, add an early bailout to it, and then use that like any other tag.

Issue #236, opened 2019-06-10
Add support for Raw string Regex


jeff-hykin:
See https://solarianprogrammer.com/2011/10/16/cpp-11-raw-strings-literals-tutorial/
```
string raw_pattern=R"((\+|-)?[[:digit:]]+)";
```
We should probably combine/cleanup the string syntax (remove c_string) before adding this. When you @matter123 added support for assembly it reminded me of this.
We can probably pull the regex patterns from the raw string regex from [python](https://github.com/MagicStack/MagicPython), and honestly I'd be okay if they stay as TextMate patterns for now so long as they are bundled up well.

matter123:
The particular syntax you posted is just a raw string (which the grammar supports). There does not appear to be a regex literal suffix

matter123:
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2006/n2053.html regex is one of the motivating examples, however. I am concerned that applying syntax highlighting for one particular format when raw string literals just as easily support html, markdown, SQL, and anything else you want may cause odd highlighting.

jeff-hykin:
Yeah you're right, whats interesting is I think python just treats all raw strings as regex haha. I guess what we should do instead is disable all the escape patterns inside the raw string.

we could possibly have a pattern that finds regex initializers `regex a_pattern(r"test")` , and then if they contain a string literal we could highlight them with regex, but it would be kind of niche. 

Its a shame there is no option to add toggle-able patterns (textmate injections), and then we could make it an option to treat all raw strings as regex literals.

matter123:
Blocked by https://github.com/Microsoft/vscode/issues/53885

matter123:
One thing that can be done while 53885 is still open, is that C++ does allow for tagged raw string literals.  
From cppreference
`prefix(optional) R "delimiter( raw_characters )delimiter"`

So if delimiter is one of:
 - `_r`
 - `re`
 - `regex`
Then the raw string would have regex highlighting.

Example:
```c++
auto re_pattern = R"re([0-9'"\+]+)re";
```

jeff-hykin:
(the merge fixed this)

PR #304, opened 2019-07-11
Tagged literal support

matter123:
Implements #236.

This PR adds a generic way of supporting tagged raw string literals. Currently, only regex patterns are supported. The regex grammar comes from embedding MagicPython's regex grammar.

raw_string literals were moved to a separate file to be able to support many different types of tagged literals without cluttering the main grammar.

Example Image:
Screenshot from 2019-07-11 01-30-52


matter123:
With the tests passing, I'm removing the WIP status.

matter123:
When #295 is merged in, we should make a copy of magic pythons regexp and add an early bailout pattern of `/(?=\)(?:_r|re|regex)")/`

That should prevent the issue described in https://github.com/MagicStack/MagicPython/issues/27#issuecomment-155483716 which is present with the PR as well.

matter123:
Based on https://github.com/microsoft/vscode/issues/1751#issuecomment-414877322
support for SQL has been added.
Screenshot from 2019-07-15 19-17-12

jeff-hykin:
So does this require that the string start with "re" for it to be regex and the string start with "SQL" to be sql?

matter123:
For regex: _r, re, regex
For sql: SQL, PSQL

jeff-hykin:
Wouldn't that mess up the regex pattern though?
Here's an example from https://solarianprogrammer.com/2011/10/16/cpp-11-raw-strings-literals-tutorial/
```
string input;
regex rr(R"(((\+|-)?[[:digit:]]+)(\.(([[:digit:]]+)?))?((e|E)((\+|-)?)[[:digit:]]+)?)");
cout << "Give me a real number!" << endl;
cin >> input;
if ( regex_match(input,rr) )  {
        cout<<"float"<<endl;
} else {
    cout<<"Invalid input"<<endl;
}
```

matter123:
Yes, that would not get syntax highlighting because it's not tagged. As discussed in #236 tagging of generic raw strings requires support from vscode.

jeff-hykin:
Wow, I need to do a better job of reading. Thanks for this feature 👍

Issue #307, opened 2019-07-12
Bad coloring of templated call to parent constructor in initialization list


kylrth:
Thanks for the quick response on my last one! Here's another one.

## To reproduce

Type this in a C++ header file (e.g. `sample.hpp`) in VS Code:

```cpp
template <typename T>
class BaseClass {
   protected:
	// no instantiation of this base class
	BaseClass(bool argOne, bool argOther) : m_one(argOne), m_other(argOther) {
	}

	bool m_one, m_other;
};

class ThisTest : public BaseClass<ThisTest> {
	ThisTest()
	    : someMember(true),  // here is a comment that is colored nicely!
	      anotherOne(false),
	      quiteALotOfMembers(false),  // here's another one!
	      soManyThatTheConstructorIsLong(true),
	      iMeanReallyLong(false),
	      BaseClass<ThisTest>(true, false) {  // this constructor call doesn't look good
	}

	bool someMember;
	bool anotherOne;
	bool quiteALotOfMembers;
	bool soManyThatTheConstructorIsLong;
	bool iMeanReallyLong;
};

```

## Expected
Templated initialization calls should be correctly highlighted.

## Actual
The Templated initialization calls are highlighted with the same color as the parentheses.

## Screenshot

![bad highlighting](https://user-images.githubusercontent.com/5044802/61150775-d4e95300-a4a9-11e9-9df9-bc8ed252080e.png)

## Configuration

Extension version: 1.12.20
C/C++ extension version: 0.24.0
VS Code version: 1.37.0-insider (user setup)
Commit: 46cc0a1670b7047e610aaef1bf84f1190802e9ec
Date: 2019-07-11T05:39:42.001Z
Node.js: 10.11.0
OS: Windows_NT x64 6.3.9600 (Windows 8)

Edit by, @matter123: corrected information in expected, actual.

PR #308, opened 2019-07-12
allow template calls in initializers, rewrite template_call_innards

matter123:
Fixes: #307
Fixes: #314


jeff-hykin:
I added subroutines to the library to enable the templates to be matched to any depth. I also took away the "remove_numbered_capture_groups" so that the inherited type can be fully matched. Once I merge in retagging we can add the "inherited" back to it

jeff-hykin:
After you've looked over it we can merge this into master

source/languages/cpp/generate.rb
```diff
@@ -724,27 +724,45 @@ def generateBlockFinder( name:"", tag_as:"", start_pattern:nil, needs_semicolon:
...
-    no_brackets_at_all = /[^<>]*/
-    balanced_brackets = /[^>]*+<[^>]*+>/
-    some_number_of_angle_brackets = oneOrMoreOf(no_brackets_at_all.or(balanced_brackets))
-    # TODO: change this to a readble form (above) once possessives (aka no back_track) is imlpemented
-    # otherwise this pattern fails because its to computationally expensive
-    some_number_of_angle_brackets = /(?:[^<>]++(?:<[^<>]++>)?)++/
+    no_anglebrackets_at_all = /[^<>]*/
+    balanced_brackets = //
+    # this is effectively what is happening (below)
+    some_number_of_angle_brackets = oneOrMoreOf(no_anglebrackets_at_all.or(balanced_brackets))
```
matter123: This should probably be a comment.

source/languages/cpp/generate.rb
```diff
@@ -3768,35 +3785,6 @@ def generateBlockFinder( name:"", tag_as:"", start_pattern:nil, needs_semicolon:
                 include: "#preprocessor_rule_define_line_context"
             }
         ]
-    legacy_memory_new_call = {
```
matter123: Why was this removed?

jeff-hykin: it wasn't used anywhere and I just realized that when I was trying to remove the `template_call.without_numbered_capture_groups`

jeff-hykin:
> Changes look good. The only slight issue is that in isolation it is not immediately clear that `rematch` performs a subexp call rather than a back reference. But I can't thnk of a better name.

Yeah I agree. I named it `recurisvelyMatch()` the first time, but its not always recursive. Then again, if its not recursive its pretty easy to just copy and paste the pattern so maybe it should be named `recurisvelyMatch()`

jeff-hykin:
`recursivelyMatch()` and `matchResultOf()` are still not great names so let me know if you think of anything better

PR #312, opened 2019-07-16
rewrite include to only be a single line

matter123:
This PR rewrites the meta_preprocessor_include pattern from a range to a simple match.

### Motivation
---
#### Standard Conformance
[[cpp.include]](eel.is/c++draft/cpp.include) Specifies that a header path cannot contain a newline. The original pattern was over matching by being a range.

Additionally, paragraph 4 permits a macro to be used as a the include path. This currently has no highlight for such contructs. 

#### Avoiding repaints
If the `#include` directive is actively being typed, and as a result is incomplete/ The current grammar would often cause repaints of large swaths of the document. This is not only slow and unnecessary but often annoying.

#### Supporting compiler extensions
See https://github.com/textmate/c.tmbundle/issues/66

### DIFF
--- a/source/languages/cpp/generate.rb
+++ b/source/languages/cpp/generate.rb
@@ -2463,54 +2463,43 @@ cpp_grammar = Grammar.new(
                 }
             ]
         }
-    cpp_grammar[:meta_preprocessor_include] = {
-            name: "meta.preprocessor.include",
-            begin: "^\\s*((#)\\s*(include(?:_next)?|import))\\b\\s*",
-            beginCaptures: {
-                "1" => {
-                    name: "keyword.control.directive.$3"
-                },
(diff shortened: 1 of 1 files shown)


matter123:
Running the tests it looks like I forgot the scope `meta.preprocessor.include` and changes the scope `string.quoted.other.lt-gt.include` to `string.quoted.lt-gt.include` (note the removal of other).

These should either be fixed or documented as changes.

jeff-hykin:
We should probably keep the originals for now

jeff-hykin:
As soon as the scopes are back I'll re-merge this.

matter123:
Yeah, I'm working on a fix with the correct scopes and support for `#include <foo> //comment`

Issue #181, opened 2019-05-20
Provide way to "re-tag" patterns.


matter123:
The currently correct way to re-tag a pattern is `.without_numbered_capture_groups` however, this only provides a way to re-tag capture group 0, if one wants to re-tag a different capture group the pattern needs to be duplicated. Additionally, as #130 shows this re-tagging can sometimes grab too much.

I was thinking something like
```ruby
partial_member.reTag(
    "variable.other.object.access" => "variable.other.object.property"
)
```
where the key would be:
 1) an existing tag
 2) a group number
 3) or a reference

and the value is the new tag.

Existing tags that are left alone can either be dropped or can be carried through.

jeff-hykin:
I think thats a good design.

When I originally worked on designing the library, I wanted a retagging feature but I couldn't think of a simple way to do it so I fell back on just wrapping them inside of functions.

Issue #290, opened 2019-07-08
Publish a textmate gem


jeff-hykin:
I'm creating this as a checklist for publishing the gem
- create a Pattern class instead of having newPattern use regex
- use the turnOffNumberedCaptureGroups on every regex pattern
- add a check for newlines, issue a warning
- switch to snake_case for everything
- check/fix the `tag_as:` in oneOrMoreOf() or zeroOrMoreOf()
- add a `to_s` method to Pattern and PatternRange
- retagging feature
- create an "export_all" which exports the language tags, json syntax, and yaml syntax to a folder
- create a standard library of patterns
    - numerics
    - block and parentheses finder
    - strings and escapes
    - inline comments
    - standard space
- add documentation to a gitbook
  - The Grammar class
    - init arguments
    - $initial_context
    - setting repos
    - exporting
    - the named patterns (@spaces, @word)
  - The Token class
    - That
      - using `not()`
    - RepresentationsThat
  - The Pattern class
    - init arguments
      - backreferences
      - $match
      - $reference
    - all of the helper methods (then, or, zeroOrMoreOf, oneOrMoreOf, etc)
  - Pattern Range
    - init arguments
  - The standard library

matter123:
We may want to reevaluate generating `*.tmLanguage.yaml` files. With, afaik, no text editor or tool supporting yaml language, and with the JSON files being pretty-printed, I not sure what their purpose is.

jeff-hykin:
yeah. I think its safe to remove them, I only added yaml because I didn't pretty print the json.

matter123:
Fairly minor concern but 
> create a Pattern class instead of having newPattern use regex

would prevent `/foo/.maybe(bar)` from working, correct?

If so, there are 33 instances of that that would need to be replaced.

jeff-hykin:
Yeah, and that was the original reason that the pattern class was never created. If I made it a class the idea would be to clean everything up and let it be symbolic, and then only convert it to regex once `to_tag` was called. But I'm not sure it's worth all the trouble.

jeff-hykin:
I've also been debating the snake case switch. I prefer static data to be sake case and functions to be camel, but Ruby uses snake for both. If I switched it would only be for formatting consistency.

matter123:
It seems the majority of the uses for starting with a regexp is looking for a literal (17), with anchors being the nxt largest (6).

matter123:
[...]

PR #322, opened 2019-07-22
Add retag method

matter123:
This PR adds the reTag feature as discussed in #181

`reTag` allows duplicating a pattern while mapping a tag name, reference, or group number to a new tag name. If the special keys "all", or "keep" are set to true, unchanged tags are kept, otherwise, they are removed. This does preclude using "all" or "keep" as key for re-tagging purposes.

`reTag` does not modify the original pattern and preserves group numbers and references.

Example Usage:
```ruby
grammar[:original] = original = newPattern(
    match: /new/.then(
        match: /\s?pattern/,
        tag_as: "entity.name.inner",
        reference: "inner"
    ).then(
        match: /\s?test/,
        reference: "bar"
    ),
    tag_as: "entity.name.outer"
)

grammar[:new] = original.reTag(
    "entity.name.inner" => "entity.name.retagged",
    "all" => true
)
```
The above example produces the following grammar fragment:
```json
"original": {
    "match": "new(\\s?pattern)(\\s?test)",
    "captures": {
    "1": {
        "name": "entity.name.inner.retag"
    }
    },
    "name": "entity.name.outer.retag"
},
"new": {
    "match": "new(\\s?pattern)(\\s?test)",
    "captures": {
    "1": {
        "name": "entity.name.retagged.retag"
    }
    },
    "name": "entity.name.outer.retag"
}
```

Resolves: #181
Part of: #290


jeff-hykin:
to differentiate the retag name from the options I changed all and keep to symbols: 
```
grammar[:new] = original.reTag(
    "entity.name.inner" => "entity.name.retagged",
    all: true,
)
```

jeff-hykin:
I kind of think that the default value for all/keep should be true, and then it can manually be set to false. I think most of the time the retagging will just be to add additional specificity to some (or all) of the tags.  I doubt there will be many times when the tags will need to be removed

matter123:
Yeah, I was just thinking that. Should be easy enough to default keep_tags to true. Somethink like `keep_tags = not arguments[:all] == false or arguments[:keep] == false`

source/textmate_tools.rb
```diff
@@ -539,7 +539,38 @@ def backReference(reference)
         new_regex.group_attributes = self.group_attributes
         return new_regex
     end
-    
+    def reTag(arguments)
+        keep_tags = arguments[:all] == true or arguments[:keep] == true
+        self_as_string = self.without_default_mode_modifiers
+        new_regex = /#{self_as_string}/
+        new_attributes = Marshal.load(Marshal.dump(self.group_attributes))
+        # this is O(N*M) and could be expensive if reTagging a bit pattern
```
matter123:
```suggestion
        # this is O(N*M) and could be expensive if reTagging a big pattern
```

source/textmate_tools.rb
```diff
@@ -539,7 +539,38 @@ def backReference(reference)
         new_regex.group_attributes = self.group_attributes
         return new_regex
     end
-    
+    def reTag(arguments)
+        keep_tags = arguments[:all] == true or arguments[:keep] == true
```
matter123:
```suggestion
        keep_tags = not (arguments[:all] == false or arguments[:keep] == false)
```

jeff-hykin:
If you want to, I think a great tweak to this would be an `:append` option. reTag(append: ".template") where every tag inside the pattern just gets additional specificity. I'd implement it but I'm not exactly sure how the reTag function works.

matter123:
That sounds like a great tweak. I'll add it. Does append apply to explicitly retagged scopes?

That is given `retag("foo" => "bar", append: "baz")` does foo become bar or bar.baz?

matter123:
The basic method for how reTag works is:
1. make a deep clone of the group attributes
2. for every group attribute:
   1. if there is no tag_as assume the group is for the refernce and preserve it
   2. for each argument:
      1. compare the key to the tag, reference and group number, if there is a match set tag_as to the value and mark the group for preserving
3. if keep_tags is false remove tag_as from the keys not marked as preserved
4. remove the preserved markings
5. construct and return the new regex

jeff-hykin:
> That sounds like a great tweak. I'll add it. Does append apply to explicitly retagged scopes?

Just saw this, the way you implemented it was the way I was thinking

jeff-hykin:
I tweaked it a bit to use a `__deep_clone__` method and I added retagging support to PatternRanges

PR #330, opened 2019-07-25
Add append option to reTag

matter123:
From: https://github.com/jeff-hykin/cpp-textmate-grammar/pull/322#issuecomment-514844216

The append option to retag adds a suffix to every scope, including explicitly set ones. The append option implies `keep: true`.

### DIFF
--- a/source/textmate_tools.rb
+++ b/source/textmate_tools.rb
@@ -589,7 +589,7 @@ class Regexp
         return new_regex
     end
     def reTag(arguments)
-        keep_tags = !(arguments[:all] == false || arguments[:keep] == false)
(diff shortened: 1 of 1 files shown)


jeff-hykin:
I'll mention here, I made that commit from `not` and `or` to `!` and `||` because the code didn't run. Generally I heavily prefer english versions but since ruby's `not`, `and`, and `or` have strange precedence I end up avoiding them to some degree.

Also thanks for adding the append option 👍

jeff-hykin:
Now we'll have to slowly start replacing all of the functions that were created as workarounds for not having a retag method

Issue #293, opened 2019-07-08
version in package.json is 1.0.0


alexr00:
Grammars are pulled into VS Code using a script that pulls other useful info from the repo, such as the commit number and the version number. The version number is taken from the package.json. Until recently, the version number in the package.json was an easy way to tell if I had a fix already. Now, the version in package.json is always.1.0.0. Is there somewhere else I should be looking for the version? 

Having it in the package.json is very convenient because many other repos do the same.

matter123:
As part of supporting other languages, each language has its own package.json. C++'s is at source/languages/cpp/package.json.

alexr00:
I'll update our script then. Thank you!

jeff-hykin:
Sorry about that, I didn't realize the package.json was being used outside of when the extension was published.

PR #333, opened 2019-07-27
add warning when there is a potential attempt to match characters after a newline.

matter123:
This PR adds a warning when it has been detected that there is a potential that a regex is attempting to match characters after a newline. This behavior from the regex is invalid as textmate grammars process a single line at a time.

As the detection looks at the regex string and not the underlying state machine, false positives are a possibility. Therefore if any pattern sets the key `:no_warn_match_after_newline?` to true, it and any pattern that includes it will skip the newline check.

Part of: #293


jeff-hykin:
Thanks for adding the gemfile too, I should've done that when I added the deepcopy dependency.

PR #335, opened 2019-07-29
Fix early bailout, line numbers, support glob patterns

matter123:
The early bailout addition assumed a match. This resulted in the reporting tools to not work.

Line numbers in reports have 1 added to them to account for the fact that in text editors number starts at 1.

The perf tools now parse their arguments as a glob pattern rather than individual files.


matter123:
Hold off on merging this. I want to fix the test coverage as well.

matter123:
Closing in favor of #338.

PR #342, opened 2019-08-05
Allow comment end paramater

### LINKED ISSUE #341: Bad coloring of final argument name in multi-line argument list
## To reproduce

Type this in a C++ header file (e.g. `sample.hpp`) in VS Code:

```cpp
#include <string>

void doThings(std::string line,
              std::string entryMap,   // a comment
              std::string entriesOne,  // another one
              std::string entriesTwo  // strange coloring here
);
```

## Expected

`entriesTwo` should be colored with the same color as the other arguments.

## Actual

`entriesTwo` is colored the same color as the types.

## Screenshot

badHighlighting

## Configuration

Extension version: 1.12.21
C/C++ extension version: 0.24.1
VS Code version: 1.37.0-insider (user setup)
Commit: 4a03f7e7eff2560d2e53c029321870514d19a2aa
Date: 2019-08-05T09:34:10.366Z
OS: Windows_NT x64 6.3.9600 (Windows 8)

matter123:
Fixes #341

Side note: there is a regression with the template keyword in issues/207.cpp. Is there any way of running npm t as a check with each PR? If could just display a diff from the master branch that would be even more helpful.


jeff-hykin:
> Fixes #341
> 
> Side note: there is a regression with the template keyword in issues/207.cpp. Is there any way of running npm t as a check with each PR? If could just display a diff from the master branch that would be even more helpful.

There is a way using circle CI or Travis CI. Basically they give you access to low-power server for free, you tell it how to install the dependencies and then what commands to run after the dependencies are installed. Tell it what branches to look at and then it will automatically run the tests.

I normally use the `npm run genall` and then the git-diff of all the changes shows everything that changed with the tests

Issue #343, opened 2019-08-06
Performance issue


jeff-hykin:
Originally added to:
https://github.com/microsoft/vscode/issues/76579

Code:
https://raw.githubusercontent.com/onqtam/doctest/master/doctest/doctest.h

@matter123  I haven't looked into the cause yet but I imagine its related to the repeated `=` or repeated `##`.

jeff-hykin:
```
#define DOCTEST_IMPLEMENT_FIXTURE(der, base, func, decorators)                                     \
    namespace {                                                                                    \
        struct der : public base                                                                   \
        {                                                                                          \
            void f();                                                                              \
        };                                                                                         \
        static void func() {                                                                       \
            der v;                                                                                 \
            v.f();                                                                                 \
        }                                                                                          \
        DOCTEST_REGISTER_FUNCTION(DOCTEST_EMPTY, func, decorators)                                 \
    }                                                                                              \
    inline DOCTEST_NOINLINE void der::f()
```

Looks like this is one of the problems, something with the macro syntax.

matter123:
This looks like there are widespread slowness with this file. Though this may not be the grammars fault. The total tim according to perf is 468ms. The wall time was 11 seconds.

<details>
<summary>Perf report</summary>

match failure
pattern name             average time    total time
source.c:3632              0.035506       10.332
source.c:3732              0.027700        5.762
source.c:3655              0.013474        0.189
source.c:3782              0.005938        0.707
source.c:3654              0.004261       10.001
source.c:572               0.004209        0.034
source.c:2306              0.004178       65.073
source.c:2307              0.003124        0.078
source.c:3801              0.003094        0.006
source.c:369               0.002918       52.923
source.c:2152              0.002916        0.213
source.c:3843              0.002472        0.005
source.c:833               0.002298       64.517
source.c:2622              0.002041       59.519
source.c:875               0.001982        0.008
source.c:370               0.001899        0.118
source.c:3928              0.001772        0.012
source.c:2257              0.001760       33.669
source.c:791               0.001673       43.932
source.c:3593              0.001601        0.243

matched, not chosen
pattern name             average time    total time
source.c:442               0.016371        0.033
source.c:791               0.016138       31.986
source.c:833               0.015329        8.231
source.c:538               0.014586        0.204
source.c:3900              0.012095        5.092
source.c:3773              0.009022        0.812
source.c:3834              0.008606        0.181
source.c:382               0.008342        3.787
[...]

Issue #324, opened 2019-07-23
Commas in macros not separating parameters properly


tristan957:
I am noticing weird behavior on the last comma in the macro definition, or maybe the previous two commas are wrong. Is TRUE supposed to be recognized as a boolean value in this context?

![image](https://user-images.githubusercontent.com/11936772/61687000-d9eaa580-ace6-11e9-9ac6-73dc6d65f352.png)

```c
#define HARVEST_REQUEST_PARAM_APPEND(string, key, value, edited, ampersand)                        \
	do {                                                                                           \
		(if (edited) {                                                                             \
			if (ampersand) {                                                                       \
				g_string_append_printf(string, "&%s=%s", #key, #value);                            \
			} else {                                                                               \
				ampersand = TRUE;                                                                  \
				g_string_append_printf(string, "%s=%s", #key, #value);                             \
			}                                                                                      \
		})                                                                                         \
	} while (0)
```

_Originally posted by @tristan957 in https://github.com/jeff-hykin/cpp-textmate-grammar/issues/311#issuecomment-514068573_

matter123:
@jeff-hykin This is apparently a valid macro function start. Do you want to try to support it?

```
#define foo(arg1\
, arg2, arg3)
```

In use: https://godbolt.org/z/ylRBsC

matter123:
@tristan957 is this a `.c` file? If so `TRUE` should be recognized as a boolean value.

tristan957:
Yes it is

matter123:
Huh, when I use C syntax highlighting. `TRUE` has the scope `constant.language.c`.
(v1.12.20)
![Screenshot from 2019-07-24 12-23-38](https://user-images.githubusercontent.com/714007/61822270-04834e00-ae0e-11e9-9aec-11508f5793de.png)

Header have by default C++ syntax highlighting, If it is not a header, however, I am not sure what the issue might be.

tristan957:
It was a `.h` file actually. Header files in my projects are C files. I can check the scopes on my end later today.

matter123:
Unfortunately without something like C/C++ extensions or [Modelines](https://marketplace.visualstudio.com/items?itemName=chrislajoie.vscode-modelines) to changes the extension on a per-file basis, all `.h` files end up with c++ syntax highlighting. This is because both C and C++ declare support for the extension `.h` and it would seem, that `c++` being later lexicographically,
 wins.

tristan957:
[...]

PR #345, opened 2019-08-08
Cpp/fix/#343

jeff-hykin:
@matter123  This is a complete rewrite of the preprocessor, I was getting it out of the way while working on #343. 

I've tested and everything looks good, but I might need to cleanup a scope name or two and some other small things if you want to take a look at it.

In particular, the test for issue #324 is something I want to look into. I fixed it (the trailing }'s ) but then it broke again, and it might be related to the single line macro.

I haven't tested this as a fix for #343 yet either, but I'm done for today.

I was hoping to merge #295 into this as well so that the preprocessor could be mostly finished.


source/languages/cpp/generate.rb
```diff
@@ -157,25 +161,14 @@ def generateBlockFinder( name:"", tag_as:"", start_pattern:nil, needs_semicolon:
 # Contexts
 #
 #
-    cpp_grammar[:ever_present_context] = [
+    grammar[:ever_present_context] = [
             # preprocessor directives, which should be a part of every scope
```
matter123: Not sure this comment is still needed.

source/languages/cpp/generate.rb
```diff
@@ -2694,7 +2426,7 @@ def generateBlockFinder( name:"", tag_as:"", start_pattern:nil, needs_semicolon:
         ),
         includes: [:function_body_context]
     )
-    cpp_grammar[:disabled] = {
+    grammar[:disabled] = {
```
matter123: Shouldn't this be part of the preprocessor file?

source/textmate_tools.rb
```diff
@@ -2,6 +2,7 @@
 require 'yaml'
 require 'set'
 require 'deep_clone' # gem install deep_clone
+require 'pathname'
```
matter123: There is no associated Gemfile change

matter123:
Per https://github.com/jeff-hykin/cpp-textmate-grammar/pull/295#issuecomment-509705461 #295 no longer modifies the preprocessor.

jeff-hykin:
I thought the branch still modified the macro context though?

matter123:
It didn't quite work right, I don't remember why, however, so I reverted that change.

source/textmate_tools.rb
jeff-hykin: I think this is a built-in library but I could be wrong, I'll check it

source/textmate_tools.rb
```diff
@@ -136,16 +137,67 @@ def recursively_set_each_value!(a_lambda)
...
-    # Globally accessible current grammar object
-    #
-    @@current_grammar = nil
-    def self.current_grammar
-        return @@current_grammar
+    # 
+    # import and export methods
+    # 
+    @@export_data
+    def self.export(*args, &block)
```
matter123: So an export always has to be followed by an import, right?
export, export, import, import won't work?

jeff-hykin: Yeah that would fail. The idea is to use Grammar.export somewhat similarly to how node.js uses module.export. 

So basically, if you want to export a textmate pattern, wrap the entire file inside of a 
```
Grammar.export() do 
end
```
And then return the things you want to export.

jeff-hykin: **However, every time `import` gets called, it should run the file, which will assign the Grammar.export. 
So `export, export, import, import` would be more like `export, export, import(export), import(export)`

source/languages/cpp/lib/preprocessor.rb
```diff
@@ -0,0 +1,362 @@
...
+            )
+            non_escaped_newline = lookBehindToAvoid(/\\/).lookAheadFor(/\n/)
+            grammar[:macro_name] = macro_name = Pattern.new(
+                match: wordBounds(identifier),
+                tag_as: "entity.name.function.preprocessor",
+            )
+        # 
+        # #pragma
+        # 
+            grammar[:pragma_mark] = Pattern.new(
```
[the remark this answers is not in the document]
jeff-hykin: That I'm not sure of, I was hoping you'd know. It shows up some of the Objective-C files, by I didn't see anything about it in the GCC standard (although I didn't look very hard).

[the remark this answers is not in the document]
matter123: It looks like it's so xcode can groups method suggestions into groups.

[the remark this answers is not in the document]
matter123: http://iosdevelopertips.com/xcode/xcode-and-pragma-mark.html

source/languages/cpp/lib/inline_comment.rb
```diff
@@ -0,0 +1,47 @@
...
+    ).then(
+        # this pattern is complicated because its optimized to never backtrack
+        match: Pattern.new(
+            tag_as: "comment.block",
+            should_fully_match: [ "thing ****/", "/* thing */", "/* thing *******/" ],
+            match: zeroOrMoreOf(
+                dont_back_track?: true,
+                match: Pattern.new(
+                    Pattern.new(
+                        /[^\*]/
```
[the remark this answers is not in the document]
jeff-hykin: yeah probably, I'll try that out and test the performance.

jeff-hykin:
oh I also wanted to mention that I couldn't get the `non_escaped_newline` to work with `$` instead of `\n` and I'm not sure why

matter123:
vscode-textmate adds a \n to the end of each line before processing it.

jeff-hykin:
but shouldnt `$`  still work for finding the end of the line?

matter123:
Yes. but the character before that is `\n` not `\`.

jeff-hykin:
oh, thanks. I thought $ matched the end of the line before the newline (even if they inserted a newline)

matter123:
It might do that in multiline mode but the patterns are processed in single-line mode so $ matches the end of the string.

jeff-hykin:
> It might do that in multiline mode but the patterns are processed in single-line mode so $ matches the end of the string.

I'll fix the `$` vs `\n` issue whenever the last major preprocessor change is made (the bailout + local context change)

jeff-hykin:
The Good news is I think the Hash pattern (instead of Pattern.new or PatternRange.new) is the :string_context_c !! So close to getting rid of nearly all the legacy patterns

PR #346, opened 2019-08-08
Cpp/fix/#343

jeff-hykin:


package.json
```diff
@@ -120,5 +120,6 @@
     "vscode-textmate": "^4.1.1",
     "yargs": "^13.2.2"
   },
-  "dependencies": {}
-}
+  "dependencies": {
```
matter123: Your fighting npm with this.

jeff-hykin: it's automatically done whenever I do npm run publish ¯\_(ツ)_/¯

PR #351, opened 2019-08-14
Contribute source.cpp.embedded.latex

matter123:
This PR modifies package.json to contribute to vscode the syntax `source.cpp.embedded.latex`.

This allows for C++ to safely be embedded in latex documents.

The example below is with modifications to Latex.plist

Before | After
-------|------
Screen Shot 2019-08-14 at 08 24 27 | Screen Shot 2019-08-14 at 08 24 16

#### Modifications to Latex.plist
In the `patterns` array for minted cpp enviorments, add the following snippet above the source.cpp dict
```xml
<dict>
	<key>include</key>
	<string>source.cpp.embedded.latex</string>
</dict>
```


matter123:
Applying the workaround/fix for the blocking issue of #352, I gave merged them together into this PR.

matter123:
With the release of LaTeX Workshop 8.0.0, this can now be tested without modification to the latex grammar.

jeff-hykin:
After thinking about this, I'm not sure that this is something we should host on this repo.

For LaTeX this means only C++ will be bailed out, even though LaTex can embed many other languages. By putting the grammar with the C++ syntax it means we have to understand and support other languages to a small degree. I think instead we should publish a tool to let maintainers create a bailed-out version of whatever grammar they need.

There is the downside that the external syntax (e.g. LaTeX won't receive changes as they happen (they will have to pull the C++ repo and re-bail it), but its probably for the best that C++ repo doesn't handle all of it's embeddings.

matter123:
The issue as stated before is that if the repository tracks the syntax file, it will dwarf many repositorys first class files.

If the extension had a release process it could clone this repository and run the prep embedded script.

matter123:
scripts/prepEmbedded.js is completely separate from everything else and could be published separately.

jeff-hykin:
> The issue as stated before is that if the repository tracks the syntax file, it will dwarf many repositorys first class files.

Would this cause some other problem? I don't see why putting the large file in the C++ repo rather than the LaTeX repo would help much

matter123:
Yeah I guess it wouldn't.

jeff-hykin:
Okay I made a repo [here](https://github.com/jeff-hykin/textmate-bailout)
I'm fine publishing it (all I have to do is run `npm publish`), but I wanted to give you the chance to clone/publish it since you wrote most of the code

matter123:
2 things of note:
1. Markdown is not the best example as it doesn't need this.
2. You should be using this branches version as it fixes while rules and avoids duplicating patters which cannot affect bailout.

matter123:
Of note in this Pr's version the bailout pattern is no longer a lookahead.

jeff-hykin:
> 2\. You should be using this branches version as it fixes while rules and avoids duplicating patters which cannot affect bailout.

Thanks, I didn't notice I copied from master. And thanks for the other notes.

npm package has been been run without errors and published

jeff-hykin:
I'll keep this merge open until I tell the LaTeX maintainer how to use it.
At that point I'll also remove the `duplicateForEmbedding` from this repo

jeff-hykin:
Okay I explained it to the maintainer

PR #352, opened 2019-08-14
Deduplicate single rule matches in embedded

matter123:
This PR modifies `duplicateForEmbedding.js` to attempt to detect patterns that cannot cause early bailout to fail. If a pattern is found it is replaced with an include to the main grammar. This reduces the latex embed by ~150k (minified).

This PR is not suitable for merging until Microsoft/vscode-textmate#105 is resolved, or this is rewritten to work around the issue.


matter123:
@jeff-hykin Do while patterns need to have their while condition modified fo early bailout to work?

jeff-hykin:
> @jeff-hykin Do while patterns need to have their while condition modified fo early bailout to work?

Thats a good question, I'm not sure. I would assume yes though since they are ranges.

matter123:
Closing in favor of #351.

Issue #353, opened 2019-08-15
Syntax highlighting for doxygen comments


matter123:
Doxygen is a popular documentation comment format for C and C++. Just as the javascript/typescript grammar has support for jsdoc comments, this grammar should have support for Doxygen comments.

Documentation comments can be distinguished by one of:
 - a block comment immediately followed by another asterisk or exclamation mark
    - (`/** ` or `/*! `)
 - a line comment immediately followed by another forward slash or exclamation mark
    - (`/// ` or `//! `)

Doxygen documentation: http://www.doxygen.nl/manual/docblocks.html

See: https://github.com/microsoft/vscode-cpptools/issues/4089

PR #354, opened 2019-08-15
Basic doxygen support

matter123:
This PR adds basic Doxygen support. Basic support means recognizing documentation comments and recognizing valid and invalid commands (both \ and @ forms).

Possible improvements:
 - recognize commands that expect a type or identifier and color the after part correctly
 - c++ highlighting in code block
 - mark \a \b \e \em sections with the appropriate markdown scope

Image:
Screen Shot 2019-08-15 at 11 07 15

Part of: #353


tristan957:
Semi-related but GTK-Doc would be a good second candidate for documentation support even if it is just something simple as recognizing the keywords similar the image in this PR.

matter123:
@tristan957 Could you link to some documentation for that?

matter123:
@tristan957 Would /[A-Z]+:/ and /@[a-z_]+:/ work?

tristan957:
@matter123 here is the documentation https://developer.gnome.org/gtk-doc-manual/unstable/documenting.html.en. I am not a pro at regex, but that looks pretty good. There is a little bit more special syntax here: https://developer.gnome.org/gtk-doc-manual/unstable/documenting_syntax.html.en. GTK-Doc also includes a way to embed C code in the second link if you scroll down to the bottom, which may be of use for highlighting within comments. But if you just wanna keep it minimal to the regex you posted earlier, I don't think anyone would complain.

matter123:
Added the above regex.
Screen Shot 2019-08-15 at 12 39 51

matter123:
It looks like the embedded support is via markdown. Makrdown support is an explicit non-goal of this PR due to upstream issues.

tristan957:
Yep I just saw that issue in the vscode repo. Looks good. Thank you for your effort.

matter123:
Added some simple extended doxygen support `\a, \b, \c, \e` and `@param` highlighting

Screen Shot 2019-08-16 at 11 07 35

matter123:
A more comprehensive Doxygen grammar would be nice and is somewhat in the works in `shared_patterns/doxygen2.rb`, but that is not a reason to block this from being merged. I'm going to add support in C for Doxygen comments, at that point it should be suitable for merging.

jeff-hykin:
Although it's pretty uncommon, I think multi-line double-slash comments `//  stuff  \ ` will screw things up (and my change didn't fix that)

fordp2002:
I think this fix has a bug where the parameter is a pointer type and prefixed with a * it does not pick it up.

Issue #356, opened 2019-08-16
Make numeric outer pattern not a range.


matter123:
Currently, the outermost portion of the numeric pattern is a PatternRange, this has a couple of disadvantages.
 - Another pattern cannot use the numeric pattern inline
 - embedded grammars need to keep the pattern despite the fact it should never cause early bailout to fail. (this adds 6-12k to the embedded grammar)

The conversion is not as simply a matter of converting the outer pattern to a match pattern, as per https://github.com/microsoft/vscode-textmate/issues/74 there is no way for a sub tokenized capture to anchor to the start of the capture. This is needed so that the list of number types fail if they cannot match the entire capture.

PR #357, opened 2019-08-16
Have numeric use an outer match pattern

matter123:
There are no regressions as a result of this change. Additionally, a bug in the parsing of numbers in `misc005.cpp` was fixed (see changed spec).

Fixes: #356


source/shared_patterns/numeric.rb
```diff
@@ -164,23 +164,28 @@ def numeric_constant(allow_user_defined_literals: false, separator:"'")
...
-            # numeric
-            binary_prefix .then(binary_digits )                        .maybe(numeric_suffix).then(binary_ending ),
-            octal_prefix  .then(octal_digits  )                        .maybe(numeric_suffix).then(octal_ending  ),
-            hex_prefix    .then(hex_digits    ).maybe(hex_exponent    ).maybe(numeric_suffix).then(hex_ending    ),
-            decimal_prefix.then(decimal_digits).maybe(decimal_exponent).maybe(numeric_suffix).then(decimal_ending),
-            # invalid
-            newPattern(
-                match: oneOrMoreOf(valid_single_character.or(valid_after_exponent)),
-                tag_as: "invalid.illegal.constant.numeric"
+            PatternRange.new(
```
jeff-hykin: Haha, very clever. I would imagine this helps maintain performance since the outside pattern is simple, and the complex inside patterns don't have to compete against all of the other patterns in an external context.

matter123: It would have negligible performance impact as the original start_pattern was a lookbehind and a lookahead, and once the PatternRange was entered it would again no longer be competing with others.

matter123: I agree that nesting helps performance, I just don't think there would be much difference between nesting with a begin/end pattern and a match/pattern.

jeff-hykin: Yeah, I meant: the alternative solution of not using any ranges. Combining all of the internal cases into one gigantic or-separated pattern.

jeff-hykin: (^ that was the first solution I had in mind for fixing #356)

matter123:
With numeric_literal being a just a `Pattern` now, it should be safe to remove preprocessor_numeric_literal, right?

jeff-hykin:
Well, its different from the regular numeric literal because it doesn't allow for user defined literals.

matter123:
While that may be the case, the preprocessor when lexing a number consumes non digits in preparation of forming a udl. Any letters stuck on the end of a number are a part of the number.

jeff-hykin:
Ah okay, then yeah we can remove it and use the normal numeric literal to help cut down on the grammar size

PR #361, opened 2019-08-20
Add linter to detect duplicate rules

matter123:
This PR adds a linter to detect when two rules only differ by their location within the grammar and are otherwise identical. These rules should probably be added to the repository.

Rules that only differ by their tag_as are not considered identical.

Example output:

```
[Duplicate]    #method_access#beginCaptures4,#member_access#captures4 are all identical rules, consider using a repository
[Duplicate]    #static_assert#beginCaptures9,#static_assert#beginCaptures4,#switch_conditional_parentheses#beginCaptures4,#switch_statement#beginCaptures5,#case_statement#beginCaptures4,#default_statement#beginCaptures4,#inline_comment#captures3 are all identical rules, consider using a repository
linting for /Users/mfosdick/cpp-textmate-grammar/syntaxes/c.tmLanguage.json: passed
```


jeff-hykin:
Where does the linter get applied?

matter123:
It is ran by `npm lint` and needs to be run separately, it works on the generated syntax files.

Issue #360, opened 2019-08-20
generate.rb should use stat to avoid regenerating grammars


matter123:
If `scripts/generate.rb` uses `stat` to determine that the source files (anything in `source` except `source/languages/`, anything in the language folder for the currently generating language) are all older than the output grammar, than expensive rebuilds could be avoided, while still updating the grammar, if needed, upon invocation of a command.

PR #368, opened 2019-08-23
Have generate use mtime to decide what grammars to generate

matter123:
This was surprisingly complicated as git commands (git pull, git checkout, etc) will set the `mtime` to the time that the command was run. This made just asking for the `mtime` usually useless.

The solution was to split the files to be tested into committed and uncommitted files. Committed files `mtime` comes from `git log -1 --format="%ad" --date=unix #{file}`, that is the time (local to the author?) that the file was last changed. Uncommitted files use `mtime` as normal

Fixes: #360


jeff-hykin:
My concern is that I do something like create a new library file (like preprocessor.rb) import it into generate.rb, but then when I run it, it doesn't generate a new grammar because it didn't detect changes.

Would that case be covered before committing preprocessor.rb?

matter123:
It  would if it was inside the source directory.

jeff-hykin:
Sounds good to me, I can merge this in next time I've got the repo pulled up. Feel free to merge it into master if you want to do it earlier 👍

PR #373, opened 2019-08-26
add missing theme and theme links

peaceshi:
can not see links and my theme in vsc. so I fix it.


peaceshi:
> Hi @peaceshi, generic_readme.md is a language agnostic template readme, while you're theme is explicitly for C++.

something in test now. I must make sure behavioral consistency between most languages.
and now also support python , lua, others (see my changlog).
for most languages , the scopes in my theme is enough , but not perfect.
this is what I should do next.

peaceshi:
> @peaceshi just as advice, you can make a [VS code extension pack](https://code.visualstudio.com/blogs/2017/03/07/extension-pack-roundup) that will install your theme, Better C++ Syntax, Lua, and all of the other extensions at the same time.

thanks! it is a good idea. I will try it soon on my dev branch.

PR #377, opened 2019-08-28
remove extraneous match rules from macro syntax

matter123:
This removes about 100kb worth of grammar rules.


jeff-hykin:
Aren't the matches necessary? don't some of the embedded ranges use the matches. Or is there some interaction I'm unaware of like repo names being searched for in parent grammars

matter123:
when `deleteMatchRules` is set to true, any include rules that reference a match rule rather than a begin/end or begin/while gets rewritten to refer to the original grammar.

You can see that [here](https://github.com/jeff-hykin/textmate-bailout/blob/master/index.js#L67)

When the bailout grammar is made at the same time as the original as is the case here, the rules that are being deleted are identical to the original rules. Only when there is a chance of version skew should the match rules be kept.

jeff-hykin:
> when `deleteMatchRules` is set to true, any include rules that reference a match rule rather than a begin/end or begin/while gets rewritten to refer to the original grammar.

Ah okay, thats awesome

Issue #415, opened 2019-11-13
C++ header with long lines hangs VSCode


matter123:
**From**: https://github.com/microsoft/vscode/issues/84723#issue-522387626
**By**: @maroc81
Issue Type: **Bug**

Loading a C++ header with long lines will cause VSCode to hang and show the "not responding" dialog.

To reproduce, create a header file with the contents such as this: https://gist.github.com/maroc81/f80bcecbd12e3b9a0496d18b970e880e

With just one line in the file, VSCode is slow but does not hang. If you duplicate the line (a hundred times or so), then trying to open the file will hang for some time. Eventually the file contents will show but VSCode is mostly unusable until you close the file.

I tested on a machine with an older version of VSCode (version 1.29.0) and it had no problems loading the file.

VS Code version: Code 1.40.0 (`86405ea`, 2019-11-06T18:14:08.920Z)
OS version: Linux x64 5.0.0-32-generic

Extensions disabled

matter123:
<details>
<summary>Perf Results</summary>
> node test/source/index.js report "perf" "/Users/mfosdick/slow.cpp"

/Users/mfosdick/slow.cpp
source.cpp
source.cpp.embedded.macro
requested grammar outside of this repository
requested grammar outside of this repository
total time: 5565.513792991638ms

match failure
pattern name             average time    total time
source.cpp:456           149.223373     4625.925
source.cpp:11866          26.565495      504.744
source.cpp:3947            0.165387        3.142
source.cpp:6895            0.084275        2.613
source.cpp:10092           0.015942        0.797
source.cpp:9994            0.014005        0.084
source.cpp:10216           0.012356        0.618
source.cpp:10791           0.011677        0.362
source.cpp:3676            0.009671        0.300
source.cpp:10134           0.009581        0.297
source.cpp:304             0.009565        0.182
source.cpp:5928            0.009548        0.296
source.cpp:7896            0.009343        0.290
source.cpp:7908            0.009308        0.289
source.cpp:5247            0.009101        0.282
source.cpp:2108            0.009052        0.281
source.cpp:6280            0.009048        0.280
source.cpp:13634           0.009043        0.280
source.cpp:12053           0.009024        0.280
source.cpp:6235            0.008997        0.279

matched, not chosen
pattern name             average time    total time
source.cpp:9977            0.034787        0.209
source.cpp:3933            0.030237        0.181
source.cpp:11424           0.019284        0.116

matched, chosen
[...]

PR #416, opened 2019-11-13
Fix for #415

matter123:
The performance issue can be dramatically
reduced by making std_space in its entirety  atomic.

This is not a complete fix.

All tests pass


matter123:
With the second set of fixes, perf reports just under 8 seconds to parse 888 copies of the bad text.

@maroc81 was it just the one line that was causing an issue?

maroc81:
We have about a half dozen header files generated from a tool with formatting similar to that bad line that suffer from this problem.  In my testing, it appears to be caused by very long text between parenthesis.

I was just able to reproduce with a different line you can try:

`int main(void /* insert 2000 spaces here*/)`

Duplicate that line a 100 times or so and you should see the same behavior.

matter123:
@maroc81 Thanks, after replacing `DEFINE_THING((` with `int main(void)` and `))` with `)` my 800 line sample file parses in 4.7s.

maroc81:
Thanks @matter123 !

jeff-hykin:
If you're confident this won't cause side effects, go ahead a merge. It'll be tomorrow night before I get a chance to look at and publish

matter123:
The only change that may cause an issue is the change of std_space from lazy to possessive. However, due to std_space being able to match nothing, I dont think its likely.

jeff-hykin:
> The only change that may cause an issue is the change of std_space from lazy to possessive. However, due to std_space being able to match nothing, I dont think its likely.

Yeah I think thats a pretty safe bet. If that does cause something to fail I imagine that individual pattern is what would need the change.

Changes published with `1.14.14`

PR #419, opened 2019-11-17
Add in and out parameters to doxygen #418

### LINKED ISSUE #418: Highlight In and Out parameters in Doxygen comments
image

Ref: http://www.doxygen.nl/manual/commands.html#cmdparam

matter123:
Resolves #418


jeff-hykin:
are spaces allowed?
```
/*!
 * @param[ out] dest
 versus
 * @param[out] dest
 */
void memcpy(void *dest, const void *src, size_t n);
```

matter123:
I haven't seen one with spaces, but I haven't looked particularly long.

Issue #412, opened 2019-11-07
Issue with comments between array declaration


Summon528:
Checklist

- [x] This problem exists even with the setting `"C_Cpp.enhancedColorization": "Disabled"`
- [x] This bug exists for C
- [x] This bug exists for C++
- [ ] This bug exists for Objective-C
- [ ] This bug exists for Objective-C++

## The code with a problem is:
```cpp
int main() {
    int a // foo
    [10]; // bar
}

```

## It looks like:
![image](https://user-images.githubusercontent.com/10328757/68364209-3d9ac080-0168-11ea-9a94-1cc1981096ea.png)

## It should look like:
// bar should be green

matter123:
Confirmed for C++ with 1.14.12

matter123:
@jeff-hykin It is treating the [10] as the start of a lambda.

I see three resolutions:
 - AFAIK lambdas cannot capture a numbers (i.e. `[10](){};` makes no sense). So a lambda capture that starts with a number cannot be a lambda capture.
 - There could be a range from type to semicolon so that [10] is never considered a lambda (part of #71)
 - Disallow lambdas at the beginning of a line in a function context.

PR #423, opened 2019-11-22
Prevent a number from starting a lambda capture

matter123:
Fixes: #412
Related: microsoft/vscode#85385


matter123:
@valentjn

This latest commit prohibits `;` or `]` (in addition to the previously added `[` from following a lambada capture.

Screen Shot 2019-11-25 at 08 47 37

matter123:
Closing as resolved in v1.14.15

valentjn:
@matter123 I don't think this is resolved. I just tested it with v1.14.16 and it only seems to resolve it when there's a number in the square brackets, not letters.

Screenshots of the above example - without Better C++ Syntax (VS Code 1.41.0, Linux x64):
Screenshot_20191230_094355

With Better C++ Syntax v1.14.16:
Screenshot_20191230_094316

The comment on line 3 is formatted correctly, but not the comment on line 10.

matter123:
Thanks, didnt catch that.

matter123:
Closing in favor of #437.

PR #431, opened 2019-12-30
Update backslash_escapes.rb

hyrious:
fix `"\018"` highlighting issue, which should be identified as two characters.
the issue occurs in c/oc/ocpp but not in cpp.


matter123:
While that was an issue that you fixed, it does not fix the issue you orignally presented. 
Screen Shot 2019-12-30 at 00 10 41
(This is `test/fixtures/pull_requests.c` as of bc92b59).

vscode-textmates `inspect` script is often useful for determining, which rule is being applied at any particular spot.

PR #434, opened 2020-01-07
Add co_await, co_yield, and co_return to the list of C++ flow control…

MansoorMohsin:
… keywords

### DIFF
--- a/syntaxes/cpp.tmLanguage.json
+++ b/syntaxes/cpp.tmLanguage.json
@@ -2105,7 +2105,7 @@
       ]
     },
     "control_flow_keywords": {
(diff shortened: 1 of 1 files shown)


matter123:
Thanks for the PR.

`cpp.tmLanguage.json` is a generated file. Could you please update the source file (`source/languages/cpp/generate.rb`) with the additional keywords. Additionally You need to create a test fixture (`test/fixtures/pull_requests/434.cpp`) and then run `npm run generate-specs`.

If all you are looking for is marking the keywords as control flow, It may be sufficent to add them to [tokens.rb#L104](https://github.com/jeff-hykin/cpp-textmate-grammar/blob/master/source/languages/cpp/tokens.rb#L104) `source/languages/cpp/tokens.rb`.

matter123:
Closing in favor of #435

PR #435, opened 2020-01-08
Add coroutine keywords

MansoorMohsin:


matter123:
Looks good, merging.

@jeff-hykin should `co_return` also have the scope `keyword.control.return`? I can imagine that some themes may highlight return separately from other control keywords. `co_return` has the same function, but I doubt that to many theme have special highlighting for `keyword.control.co_return`.

jeff-hykin:
> @jeff-hykin should `co_return` also have the scope `keyword.control.return`?

Yeah I think thats a good scope. For additional specificity we could do `keyword.control.return.co_return`

jeff-hykin:
Also thanks for merging this in

PR #436, opened 2020-01-09
Rewrite assembly pattern

### LINKED ISSUE #433: C++ asm statement not recognized
From @RSDuck in https://github.com/microsoft/vscode/issues/87991

Checklist

- [x] This bug exists for C
- [x] This bug exists for C++

```c++
void test()
{
  asm
  (
    : : :
  );
}
```

## It looks like:
See that `asm` doesn't have the right scope.

image

## It should look like:
Move the opening paren after `asm` to the same line and it gets the right scope.
image

matter123:
The detection of the mandatory parens
has been imporved to detect an opening
paren that is not on the same line as asm

Fixes #433


jeff-hykin:
thanks for the change

Whats the addition of the symbolic names?

matter123:
See section 6.47.2.5 (about 3/5 down) in https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html

Not handling them caused syntax issues furthur down.

`[prevTileLo]` in line 16 of test/issues/433.cpp

jeff-hykin:
👍

PR #453, opened 2020-02-27
Fix bad bounds when using raw alternation in pthread_types

### LINKED ISSUE #452: Functions prefixed with "pthread_t" get wrong scope
I'm seeing the `pthread_t` part of` pthread_threadid_np();` get scoped as `support.type.built-in.posix-reserved.pthread.cpp`

From @foreignmeloman in https://github.com/microsoft/vscode/issues/91529

Seems like VS Code does not highlight correctly function calls which have "pthread_t" prefix in the name:

Examples:
```C++
pthread_threadid_np();
pthread_tfoo()
```

Screenshot:
image

matter123:
Fixes #452


matter123:
The test failure looks like this fixed a hidden bug attribute scoping. attributes are not qualified types.

jeff-hykin:
Thanks 👍

PR #466, opened 2020-04-04
Fix missing reference to function_type

asottile:
as far as I can tell this has always been missing (since 7f10704e4d2ef69626fe95d493260706099de671)

here's the crash for me:

```pycon
babi/highlight.py", line 563, in _include
    return self._patterns(grammar, (repository[s[1:]],))
  File "/home/asottile/workspace/babi-grammars/venv/lib/python3.6/site-packages/babi/fdict.py", line 44, in __getitem__
    raise KeyError(key)
KeyError: 'function_type'
```


matter123:
Looking back at #135 it seems that `function_type` was for functions inside of templates, which have for the form of `A(B, C)`, while `function_pointer` matches constructs with the form `A (*B)(C)`. As such `function_pointer` is not an acceptable replacement for `function_type`.

asottile:
ah -- alternatively should I just remove `function_type` for now?

matter123:
Yeah, that seems to be the best option.

asottile:
awesome, should be good now :)

rebasing / merging in this repo is a bit annoying with the embedded grammar conflicting every time

matter123:
Could you describe how the embedded grammar conflicts when you merge/rebase? Assuming you generate before committing (which it seems you do), the embedded grammar would be overwriiten removing any merge conflict markers.

asottile:
I was merging this PR and #465 together, both of which caused a regen of the embedded grammar which is all on one line

jeff-hykin:
Yeah it is annoying. Durning the merge I've gotten used to generating the syntax and then committing that. It would be nice to have it be automatic. 

I'm comfortable with git, but I don't know many of the fancy features. There might be some kind of attribute that can be applied similar to git GLFS where it never merges and always just pics the latest timestamp

PR #467, opened 2020-04-09
fix #428

### LINKED ISSUE #428: __attribute__ space between closing  brackets breaks syntax highlighting
Checklist

- [ ] This bug exists for C
- [x] This bug exists for C++

```
#include <unistd.h>
#include <fcntl.h>

struct fff
{
  int start;
} __attribute__(( packed ) );

int main() {
  puts("aaa");
}
void foo() {
    
}
```

## It looks like:
** screenshot + theme name preferable **
image

## It should look like:
image

here I removed space

Yanpas:


jeff-hykin:
(merging into `cpp/fix/#428` to run tests)

jeff-hykin:
tests passed, fix pushed to master and published with `v1.14.29`

PR #472, opened 2020-04-28
support for syntax highlight of GLSL shaders

DoDoENT:
This allows for syntax highlight of GLSL shaders within multi-line C++11 string literals. This is very useful for easier editing of inline shaders in C++ source code.

Also added theme_demo3.cpp file which adds demonstration for both existing SQL and regex literals, as well as newly added GLSL literals.


DoDoENT:
Wow, that was quick!

Thank you!

jeff-hykin:
No problem! This was a pretty small change, and the example code helped.

@DoDoENT  I also updated the `CONTRIBUTING.md` which should explain PR's a bit better now too.

PR #490, opened 2020-07-14
add final and override to constructors

### LINKED ISSUE #489: Incorrect/inconsistent highlighting of final/override
Checklist

- [ ] This bug exists for C
- [x] This bug exists for C++

[Full live code](https://godbolt.org/z/35dsoa)

## It looks like:
[Screenshot]

## It should look like:
Both `override` and `final` keywords should be highlighted on all cases.

matter123:
Fixes #489.

The test suite is broken, I will invertigate why later.


jeff-hykin:
I fixed (kinda) the problem with the test suite. I'm just setting a private member variable to what its supposed to be since I don't know the normal method for doing so (can't find the constructor)

PR #521, opened 2020-11-13
remove unused comments_context

asottile:
I couldn't get the ruby scripts to work, they segfault on my machine 😭 so I manually made the changes to the individual outputs


jeff-hykin:
@asottile I added a [new setup method and instructions](https://github.com/jeff-hykin/cpp-textmate-grammar/blob/master/documentation/SETUP.md). Inbetween the ruby gems, node native modules, and node-gyp requiring python and C++ compilers I know its a real pain to setup, so the new setup method should automate everything other then WSL (but that's only needed if you're on Windows)

jeff-hykin:
and thanks for merging this @matter123

asottile:
yeah the weird thing is I have all the tools available, but `ruby` itself was segfaulting :S

jeff-hykin:
yeah ¯\_(ツ)_/¯ the new managed environment downloads the exact version (literally checks a SHA256 hash to make sure every file in the whole library and every dependency of the library are the same) so it makes machine-specific errors really rare and easy to debug.

jeff-hykin:
and the segfault is because ruby can import c code, and probably is doing that for one of the libraries

PR #554, opened 2021-06-07
Improve Objective-C and Objective-C++ highlighting

### LINKED ISSUE #538: Objective c: self, super, and _cmd are not syntax highlighted at all
Checklist

- [ ] This bug exists for C
- [ ] This bug exists for C++
- [x] This bug exists for Objective-C
- [x] This bug exists for Objective-C++

OS version: Darwin x64 19.5.0

I noticed that in the Objective-C language, special variables (`#special_variables`) like `self`, `super`, and `_cmd` are not syntax highlighted at all. I also noticed that method calls are not syntax highlighted either, unlike regular C-style function calls.

After some digging around in the `.tmLanguage` file for Objc, I noticed that if I add a top-level include for the `#special_variables` repository and copy the declaration for `meta.function-call.objc` to the top level as well, both are syntax highlighted.

The default rendering:
Original

With a duplicate top-level decl for `meta.function-call.objc`:
With special_variables

With the above, and `#special_variables` included:
Screenshot 2021-02-13 at 4 31 14 AM

Originally from @NSExceptional in https://github.com/microsoft/vscode/issues/116612

### LINKED ISSUE #550: The in keyword in the for-in loop syntax for Objective-c is not scoped as a keyword
Checklist

- [ ] This bug exists for C
- [ ] This bug exists for C++
- [x] This bug exists for Objective-C

In Visual Studio Code, the in keyword is not recognized as a keyword.
It is not color coded as the same color as other keywords like for, if, and return.

They keyword "in" should be color coded the same color as these other keywords.

```
    NSDictionary* myDictionary = @{
            @"myKey": @"myValue"
    };
    for (NSString* key in myDictionary) {
        
    }
```

## It looks like:
The scope list for `in` is:
meta.parens.objc
source.objc

## It should look like:
I would expect the scope list to include `keyword` at some point.

Originally from @TylerBreau in https://github.com/microsoft/vscode/issues/125171

DavidGoldman:
- Highlight `in` keyword
- Improve `self/super/_cmd` highlighting, previously
  it would trigger in an @implementation and a few other places
- Improve detection of Objective-C method calls, previously it would
  get mixed up with `meta.bracket.square.access`

Fixes #550 and fixes #538

### DIFF
--- a/source/languages/objc/examples/examples.spec.yaml
+++ b/source/languages/objc/examples/examples.spec.yaml
@@ -72,29 +72,29 @@
     - keyword.operator.assignment
 - source: '['
   scopesBegin:
-    - meta.bracket.square.access
+    - meta.bracketed
   scopes:
-    - punctuation.definition.begin.bracket.square
+    - punctuation.section.scope.begin
 - source: 'AVCaptureDeviceDiscoverySession '
 - source: '        discoverySessionWithDeviceTypes:@'
 - source: '['
   scopesBegin:
-    - meta.bracket.square.access
+    - meta.bracketed
   scopes:
-    - punctuation.definition.begin.bracket.square
+    - punctuation.section.scope.begin
 - source: AVCaptureDeviceTypeBuiltInWideAngleCamera
 - source: ']'
   scopes:
-    - punctuation.definition.end.bracket.square
+    - punctuation.section.scope.end
   scopesEnd:
-    - meta.bracket.square.access
+    - meta.bracketed
 - source: '        mediaType:AVMediaTypeVideo '
 - source: '        position:AVCaptureDevicePositionUnspecified'
 - source: ']'
   scopes:
-    - punctuation.definition.end.bracket.square
+    - punctuation.section.scope.end
   scopesEnd:
-    - meta.bracket.square.access
+    - meta.bracketed
 - source: ;
   scopes:
     - punctuation.terminator.statement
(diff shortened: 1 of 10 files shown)


jeff-hykin:
I worked yesterday on modularizing the libraries used here (pulled all the JavaScript out, reorganized it, and published it on npm)

I'm going to try to make a separate repo for each language instead of this giant multi-language monorepo. So hopefully that'll be done in the next week or so.

Ideally modifications would be in Ruby (these current changes are fine). So whenever I make the new repo, I'll give a quick demo of how to do that.

DavidGoldman:
SGTM, just let me know what I should do, thanks

jeff-hykin:
Will do, and thank you for the contributions and patience!

NSExceptional:
@jeff-hykin What do you think the timeline on that will be?

VS Code relies on these (as I'm sure you already know) and won't update their own grammars until they're updated here. I'm not in a *crazy* hurry to use this new grammar, but if you think it might be another few weeks or longer, is there any harm in just merging this now?

jeff-hykin:
This is a legitimate question, as this repo itself was made precisely because the original maintainers couldn't get changes merged in fast enough (years).

Two things
- Yes there is (some) large potential for downside. Adding one bad pattern can totally wreck many peoples highlighting, and they won't know how to fix it
- This repo is both an extension and a VS Code upstream source. The extension is "bleeding edge" meaning yes I should merge first, test later and this PR deserves that. However, the VS Code source is the opposite; it can take half a year for changes in this codebase to show up in VS Code even if I merged it today


I can merge the code, and comment on a PR while on the elevator, or waiting for a bus. But it's not going to show up in the extension until I get to my desktop setup with the intention of opening up the codebase, logging into vsce, packaging, and publishing. Which I think is what you're wanting.

NSExceptional:
Pardon my confusion, but which extension are you referring to? VS Code bundles these grammars into the app itself, right? I can see the files inside the `.app`

Or is it one of those "built-in extension" type things?

jeff-hykin:
People with the extension are beta testers, after changes have been data tested for quite a while without major problems, they get merged into VS code official. 

The extension is at the top of the readme.md 
https://marketplace.visualstudio.com/items?itemName=jeff-hykin.better-cpp-syntax

It's not a built in extension

NSExceptional:
Ohhh I see, gotcha.

Do you ping someone at Microsoft or open a PR whenever you decide the beta test for a particular change is over?

jeff-hykin:
> Do you ping someone at Microsoft or open a PR whenever you decide the beta test for a particular change is over?

I normally only ping them if theres a critical bug that got fixed, otherwise they test and pull in updates automatically over time.

jeff-hykin:
@DavidGoldman  Alright, that was a little longer than expected but I finished overhauling this repo. It has now been properly broken up into 4 repos.

- [better-objc-syntax](https://github.com/jeff-hykin/better-objc-syntax)
- [better-objcpp-syntax](https://github.com/jeff-hykin/better-objcpp-syntax)
- [better-c-syntax](https://github.com/jeff-hykin/better-c-syntax)
- [better-cpp-syntax](https://github.com/jeff-hykin/better-cpp-syntax)

Now it should be easier to get stuff merged in, and the codebase should be easier to understand/fork/run. There really hasn't been any work done on Objective C/C++ since we seperated it from the C++ syntax several years ago so I'm sure lots of improvements are possible.

Issue #396, opened 2019-10-11
C++20 keywords don't seem to be highlighted


cjdb:
Checklist

- [x] This problem exists even with the setting `"C_Cpp.enhancedColorization": "Disabled"`
- [ ] This bug exists for C
- [x] This bug exists for C++
- [ ] This bug exists for Objective-C
- [ ] This bug exists for Objective-C++

## The code with a problem is:
```cpp
module example;
import <vector>
import some.other.module;

export template<typename From, typename To>
concept convertible_to =
	std::is_convertible_v<From, To> and
	requires(From (&f)()) {
		static_cast<To>(f());
	};
```

In the above example, `module`, `import`, `<vector>`, `some.other.module`, `concept`, and `requires` should all be highlighted, but they're not.

## It looks like:

![image](https://user-images.githubusercontent.com/8622748/66616383-3a212180-ec1c-11e9-877b-fbed834d64df.png)

**Theme:** peaceshi

![image](https://user-images.githubusercontent.com/8622748/66616429-73599180-ec1c-11e9-84b6-c7c127063cd4.png)

**Theme:** XD 

![image](https://user-images.githubusercontent.com/8622748/66617281-38f1f380-ec20-11e9-8c4f-1e72962d899d.png)

**Theme:** One Monokai

## It should look like:
* `module`, `import`, `concept`, and `requires` should all be highlighted like the other keywords. (Note: `module` is contextually-dependent.)
* `<vector>` and `some.other.module` should probably look like `<vector>` in `#include <vector>`

cjdb:
I'd be okay with `some.other.module` not being highlighted.

matter123:
Duplicate of #302. Current focus is on published standards.

matter123:
Just keyword highlighting, no context awareness, could be added in easily enough. Import as an alternative to` #include` should also be easy enough to add.

cjdb:
I've been trying to work out how to do this locally, but without success. Looks like the above words are already in the JSON file, which has confused me.

matter123:
Often those are in negative lookbehinds and lookaheads.

`import`: the `import` keyword needs to be pulled out of the check for `#`, But if the semicolon is required it might be best if its pulled into its own pattern.

Adding keywords: This should be just as simple as adding the correct flags to the keywords in https://github.com/jeff-hykin/cpp-textmate-grammar/blob/master/source/languages/cpp/tokens.rb#L261-L274

[...]

PR #574, opened 2021-10-14
Add consteval keyword highlighted. (fixes #573)

### LINKED ISSUE #573: C++20 consteval keyword is not highlighted
This is a new keyword in C++20. 

Some example usages can be found at https://en.cppreference.com/w/cpp/language/consteval. Currently, none of them are highlighted (scope is just `source.cpp`).

soroshsabz:
ITNOA

I try to add consteval keyword highlight. this PR is resolved #573


language_examples/feature_consteval.spec.yaml
```diff
@@ -0,0 +1,122 @@
...
+- source: " ITNOA\r"
+- source: //
+  scopes:
+    - punctuation.definition.comment
+- source: " consteval is not coloerize, so it is wrong\r"
+  scopesEnd:
+    - comment.line.double-slash
+- source: constexpr
+  scopesBegin:
+    - meta.function.definition
```
soroshsabz, line 15: Why constexpr has scopesBegin, but cosnteval does not have scopesBegin?

@matter123

soroshsabz:
@matter123  @jeff-hykin Please review it

thanks :)

jeff-hykin:
looks good! Next time I have the repo open I'll publish it. And just post a bump message if it's been a week or two and still isn't in the Extension

language_examples/feature_consteval.spec.yaml
soroshsabz: @jeff-hykin can you explain to me, why this happen?

jeff-hykin: I don't know.

`meta.function.definiton` Just means the start of a function. It should stop at the `}`. 

Then it should start again with `consteval`

But it never stops.

soroshsabz: I thinking same, but as you can see with `constexpr`, `meta.function.definito` is begin but in `consteval` I could not found `meta.function.definito` in source of `consteval`

I think it is like bug, because I think `constexpr` and `consteval` is behave same and fully similar in function declaration.

soroshsabz:
related to #396

PR #593, opened 2022-06-27
Better handling of a raw string as a param

### LINKED ISSUE #530: A raw string literal inside a struct/class definition breaks the highlighting
Checklist

- [ ] This problem exists even with the setting `"C_Cpp.enhancedColorization": "Disabled"` → Not sure, see below
- [ ] This bug exists for C
- [x] This bug exists for C++

The issue persists even after I add `"C_Cpp.enhancedColorization": "Disabled"` to settings.json and restart VS Code, but VS Code says that `C_Cpp.enhancedColorization` is an unknown configuration setting, so I am not sure if I am adding the setting correctly. I am using VS Code 1.51.1 on Mac.

```
struct S {
  int x =
      f(R"(some raw string literal)");
  int y = 100;
  const char* z = "hello";
};
```

## It looks like:
image

Color Theme: Default Light+

## It should look like:
The raw string literal should be colored like a string literal, and everything after that should not be colored like a string literal.

a-stewart:
Rather than defining the string_context as a child of the function_parameter_context, we seem to get a more reliable result if we define it as part of the parameter tag (`meta.parameter.cpp`).

This should fix #530 in more situations.


jeff-hykin:
Thanks! Sorry it took so long to merge

PR #594, opened 2022-07-03
Cpp/fix/#452

j-cortial:
This PR is my attempt to fix issue https://github.com/jeff-hykin/better-cpp-syntax/issues/427.


jeff-hykin:
Thanks for the contribution! it looks great

jeff-hykin:
Your fix should be available on `v1.15.19` of the extension now

PR #595, opened 2022-07-04
Cpp/fix/#407

j-cortial:
Here is my proposal for fixing issue https://github.com/jeff-hykin/better-cpp-syntax/issues/407 (see also duplicate issue https://github.com/jeff-hykin/better-cpp-syntax/issues/496 and related issue https://github.com/jeff-hykin/better-cpp-syntax/issues/587).

I introduced four additional TextMate scopes, namely `keyword.other.[default|delete].[destructor|function]` since any special member function can be defaulted (not only constructors) and any function can be deleted. I did not remove the existing scopes `keyword.other.[default|delete].constructor` for destructors to ensure backward compatibility.


jeff-hykin:
Thanks for the fix! Sorry it took so long to merge. Its now pushed on `v1.15.21`

PR #596, opened 2022-07-10
Cpp/fix/#492

j-cortial:
This pull request should take care of issue https://github.com/jeff-hykin/better-cpp-syntax/issues/492. However it will probably not work for more involved cases. More work is needed for that, see issue https://github.com/jeff-hykin/better-cpp-syntax/issues/246.


jeff-hykin:
Just wanted to say, I'll be looking at and probably merging all your PR's next time I'm in the codebase. I appreciate all the fixes!

PR #599, opened 2022-07-22
Fix typos in README.md

jogo-:


jeff-hykin:
Spelling never was my strong suit 😅, can't believe those have been sitting there for so long. Thanks for the contribution

PR #609, opened 2022-10-23
Cpp/fix/type alias

j-cortial:
Narrow definition of type alias, only an naked identifier can be on the LHS of the assignment.


j-cortial:
Narrow definition of type alias

PR #612, opened 2022-11-02
Cpp/fix/#610

j-cortial:
This PR fixes the sample code in https://github.com/jeff-hykin/better-cpp-syntax/issues/610 but is not a complete solution.
I would like to have your opinion on the fix before I go ahead and tackle raw string and character literals.


j-cortial:
Now raw strings and (single-quoted) characters should be covered as well as standard double-quoted strings.

I did not use a scope beginning with `keyword.other.unit.suffix` since most suffixes (and especially string suffixes) have nothing to do with (physical) units. 

The scope I used is `keyword.other.suffix.literal.user-defined` (to match the term used by https://en.cppreference.com/w/cpp/language/user_literal) and the refinement `keyword.other.suffix.literal.user-defined.reserved` when the suffix is reserved by the implementation (does not start with an underscore, such as the standard `s` and `sv`).

jeff-hykin:
Thanks again! Published in `v1.17.0`

PR #622, opened 2023-01-04
Make suffix tags more consistent

j-cortial:


j-cortial:
https://github.com/jeff-hykin/better-cpp-syntax/issues/610

PR #680, opened 2025-04-04
Pack indexing

### LINKED ISSUE #677: Pack indexing breaks syntax highlighting
Checklist

- [x] This problem exists even with the setting `"C_Cpp.enhancedColorization": "Disabled"`

```cpp
template <typename... T>
constexpr auto first_plus_last(T... values) -> T...[0] {
    return T...[0](values...[0] + values...[sizeof...(values)-1]);
}
int main() {
    // first_plus_last(); // ill formed
    static_assert(first_plus_last(1, 2, 10) == 11);
}
```

(from [P2662R3](https://wg21.link/P2662R3), the paper that introduced [pack indexing](https://en.cppreference.com/w/cpp/language/pack_indexing))

## It looks like:

Image

(Dark+)

## It should look like:
`static_assert` should be blue, not yellow.

seishun:
Fixes https://github.com/jeff-hykin/better-cpp-syntax/issues/677.

I'm not happy with the `->` being mistagged, but I'm also not excited about fixing it for [pack-index-specifier](https://eel.is/c++draft/dcl.type.pack.index#nt:pack-index-specifier) while keeping it broken for [decltype-specifier](https://eel.is/c++draft/dcl.type.decltype#nt:decltype-specifier) (see [computed-type-specifier](https://eel.is/c++draft/dcl.type.simple#nt:computed-type-specifier), [simple-type-specifier](https://eel.is/c++draft/dcl.type.simple#nt:simple-type-specifier), [type-specifier](https://eel.is/c++draft/dcl.type.general#nt:type-specifier), [type-specifier-seq](https://eel.is/c++draft/dcl.type.general#nt:type-specifier-seq), [type-id](https://eel.is/c++draft/dcl.name#nt:type-id), [trailing-return-type](https://eel.is/c++draft/dcl.decl.general#nt:trailing-return-type)), and I'm not sure if trying to update the pattern in `qualified_type` is the right approach.

Also I'll need help with scope naming because I don't understand the naming scheme.

### DIFF
--- a/language_examples/pack_indexing.cpp
+++ b/language_examples/pack_indexing.cpp
@@ -1,0 +1,4 @@
+template <typename... T>
+constexpr auto first_plus_last(T... values) -> T...[0] {
+    return T...[0](values...[0] + values...[sizeof...(values)-1]);
+}
--- a/language_examples/pack_indexing.spec.yaml
+++ b/language_examples/pack_indexing.spec.yaml
@@ -1,0 +1,140 @@
+- source: template
+  scopes:
+    - storage.type.template
+- source: <
+  scopes:
+    - punctuation.section.angle-brackets.begin.template.definition
+- source: typename
+  scopesBegin:
+    - meta.template.definition
+  scopes:
+    - storage.type.template.argument.typename
+- source: ...
+  scopes:
+    - punctuation.vararg-ellipses.template.definition
+- source: T
+  scopes:
+    - entity.name.type.template
+  scopesEnd:
+    - meta.template.definition
+- source: '>'
+  scopes:
+    - punctuation.section.angle-brackets.end.template.definition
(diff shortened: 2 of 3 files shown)


jeff-hykin:
> I'm not sure if trying to update the pattern in qualified_type is the right approach

It is, but this PR is a good starting point cause its a lot easier to understand.

And yeah dont worry about the scope naming, theyre all just made up names with some very rough/vague patterns.

seishun:
How about `meta.pack.indexing` for the whole expression, `variable.other.pack` for the pack (even though it's not a variable in any meaningful sense...) and `punctuation.definition.begin.bracket.square.pack.indexing` and `punctuation.definition.end.bracket.square.pack.indexing` for the opening and closing bracket respectively?

PR #707, opened 2026-05-24
Fix multi-line block comments in preprocessor macros

dnicolson:


jeff-hykin:
Thanks for contributing! I'll try to get this published.

PR #708, opened 2026-05-31
Add highlighting for C++23 universal character names

eisenwave:
C++23 got `\u{A0}` as an alternative to `\u00A0` as well as `\N{NO-BREAK SPACE}`. These should have highlighting.


eisenwave:
@jeff-hykin ?

PR #711, opened 2026-07-17
Add %b specifier support (fix #700)

### LINKED ISSUE #700: C/C++ linter does not support "%b" specifier in printf-formatted strings
The linter (syntax coloring) for formatted strings displays an "error" (red character) for the %b specifier in C, even though this specifier is now supported since C23

Steps to reproduce:
1. `touch printf_linter_bug.c && code printf_linter_bug.c`
2. Paste:
```c
int main(){
	printf("%b \n", 0x7f);
	return 0;
}
```

We observe that the '%' is coloured in red and the b in the string colour, whereas a correctly supported identifier (e.g. "%d") would show both in blue

This is valid C23 code, which you can check by compiling (with an up to date compiler): `clang -std=c23 printf_linter_bug.c -o printf_linter_bug && ./printf_linter_bug` (the command succeeds)

Magyar57:


Magyar57:
I implemented it for the C++, but I want to do it for C as well. Do I need to modify better-c-syntax for that ? I can't find the same regex that I had to modify here

